PR #9, opened 2022-06-18
Fenbranch

Tk420634:
Please double check this Andy.  It should add the back door to the vault and overhaul the NCR base into the first take of the RRAD dungeon.  But I am almost *certain* I have fucked something up badly.


DameonOwen:
Yes, I think you did. Sunset-Lower says it was changed to use /turf/open/floor/wood instead of /turf/open/floor/wood_common, but that path does not exist in the code (as far as I can see). Also, /datum/map_template/shuttle/enclave_elevator exists on that map and it blows up on compile.

DameonOwen:
Yep, that is the only big thing that I noticed, that path change on Sunset and Sunset-Lower.

GremlingSS (approved):
Seems Gud
Yeah my wood floor changes got overwritten but I shouldnt have pushed those changes in the first place.

PR #11, opened 2022-06-19
Headset changes!

GremlingSS:
## About The Pull Request
Forces vault headsets to be set to the vault frequency, stops you from accessing public common by accident.

Vault intercomm will also be given the same treatment~

Was ordered to also include an edit to genitals which reenables missing genitals

Fixed a lot of runtimes happening too, though it makes the map look ugly in some areas, but it's better than having no turf!

## Why It's Good For The Game
yes.

## Pre-Merge Checklist

## Changelog

:cl:
tweak: Made vault radios only work in vault frequency
add: genitalia
/:cl:


Tk420634 (approved):
Jon and Faire are stinky, but these should be good.

PR #22, opened 2022-06-22
Window Name and Lobby Music

Tk420634:
Changes the window from Sunset Wasteland to Coyote Bayou.

Adds some more lobby music, most of which is era appropriate though one, Cotton Fields, is only added because it references Texarkana.


GremlingSS (approved):
seemsgood

Tk420634:
dumbass tk moment

PR #26, opened 2022-06-22
Better title screen

Tk420634:
Greetin's n' meetin's.


GremlingSS (approved):
yee, one issue being size diff, make sure it works, because of this, you do the honors once you double checked it fam

Tk420634:
Oops, didn't realize size was futzed.  I'll fix in a moment.

PR #34, opened 2022-06-24
Mapping stuff

thingpony:
Pushing what I have cause I  can't be assed.


Tk420634 (approved):
Adds some stuff to Nash, and the tribal medtent.

PR #35, opened 2022-06-24
Update Pahrump-Sunset.dmm

thingpony:
Makes the strip club into a proper bar and diner, part one. Needs actual booths and tables.
Adds a road through where Nash will be
Removes Bighorn's bar and the Khan farm thing.
Adds an area to the medical tent in tribal camp
Adds a dirty bed for recovery or to give your patient infections. Either one.

### DIFF
--- a/_maps/map_files/Pahrump-Sunset/Pahrump-Sunset.dmm
+++ b/_maps/map_files/Pahrump-Sunset/Pahrump-Sunset.dmm
@@ -150,20 +150,6 @@
 	icon_state = "yellowrustysolid"
 	},
 /area/f13/building/massfusion)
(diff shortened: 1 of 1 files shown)


Tk420634 (approved):
Nash now WIP

GremlingSS (approved):
Only concern I have is full on removals of certain tiles, which might conflict with Tk's, but I doubt TK uses a tile which as has a toilet, light, dirt, vomit and a turf which uses "bluerustychess2" as it's iconstate and also has the bighorn area, but that might cause conflicts on his end if he has used that, thats where conflicts happen now with mapping is if you remove a certain tile he uses.

PR #37, opened 2022-06-24
Dungeon removal, and some small map changes

Tk420634:
Removed Sunsets dungeons, they can still be found in the older map files folder for future reference.  For now they were just taking up space, and don't fit with what we want moving forward.

Also removed their dungeon elevators from the main map.

Dungeon Z was overhauled entirely, setting it up for moving forward with an area for the Texarkana Mall.  The Enclave Bunker/BOS Bunkers are placed down with future plans to make them easy to setup for events.

Also also added a few things to the vault, and poked about in it to straighten some stuff out.  Mostly added just a few things to the armory.


GremlingSS:
awww one conflict, lets see....

Tk420634:
Might be the tribal medtent.  I think we both added its area.

GremlingSS (approved):
early approval, just wait for linter and whatnot finish successfully first before merging <3

PR #38, opened 2022-06-24
stops lights from flickering

GremlingSS:
https://youtu.be/EGJJf6ME7KI


Tk420634:
why that youtube video tho?

Tk420634 (approved):
Kill them with fire, hopefully this fixes the occasional gigantic lag spikes.

PR #43, opened 2022-06-25
Blurbies

GremlingSS:
## About The Pull Request
Ooop, called the commitment wrong.
RIGHT.
I made an oopsie and didnt actually make mobs move the way they should with the new movement system.
I made it so speech bubbles will make a speaking sound and will loop until you're done. 
allowed the option for it to play after
or when you start typing.

## Pre-Merge Checklist


Tk420634 (approved):
hero pr, madman supreme

PR #48, opened 2022-06-26
Followers Clinic Update

Tk420634:
-Adds real Defibs, clothing machines, more prosthetic limbs and proper access restricted front doors.


GremlingSS (approved):
image

PR #52, opened 2022-06-26
Map Overhauls

Tk420634:
Added two more back and forth entrances for people to get to the Redwater Z from Nash.

Overhauled Redwater Z's top left quarter to be a proper fucking swamp.

Feel free to add some buildings and stuff in it!


Tk420634:
Missed something

PR #53, opened 2022-06-26
Dumbass Moment - Map Overhauls

Tk420634:
Check the two files for their changes.  Mostly just Redwater map changes, but also some rechargers in the Followers Clinic for IPCs


GremlingSS (approved):
yes

PR #60, opened 2022-06-28
Removed rocksprings gates for now

Tk420634:
I'll readd them later, but honestly the town can be open.


Tk420634:
Alright, this is good now.

> a

It'll be good good here in a sec.  God damn inside water tiles n shiet man i stg

PR #77, opened 2022-06-30
Preliminary Tribal Faction Setup

Tk420634:
This gets the Tribal faction, the Sulphur Bottom Tribe, into a semi-working state, with spawn points and four loadouts.

The issue with the old codes tribals is that whoever took them out cut bits of code out for some reason, and I'll need help turning them back on properly.  For now tho, these tribals work great for an 'early settler' group to the area.  I've written it into their lore description that they're very early to the area, and will pick a chief later on.


Tk420634:
Oops.

PR #100, opened 2022-07-01
Adds the hellpig in.

thingpony:
Just set the damage to 15 and 25. No idea how strong this thing is so, can always adjust later.


thingpony:
Credit goes too SlapDrink#0083

PR #107, opened 2022-07-02
Excess fat removal - Sound Aurora Caelus

Tk420634:
idgaf about removing the lisence for this song, a few bytes of data for text means nothing to me but a 4mb worthless sound file is getting yeeted.


Tk420634:
Fuck you, checks.  These are all worthless sound files.  I'm merging.

PR #112, opened 2022-07-02
Heavy Music File Removal (TEST)

Tk420634:
These files aren't referenced anywhere proper, but they do play and cause issues.  I'm really flabbergasted by this.  I assume they're being called somewhere in a config file of some sort?  Maybe?  I just don't know.  But the game compiles and runs without runtimes with them removed, so I say chuck them in the trash to improve performance.


Tk420634:
Closing to roll back and find where I fugged up

PR #273, opened 2022-07-14
Buffs plasma bolts

Superlagg:
## About The Pull Request

Buffs plasma bolts quite a bit. The pistol was too weak, worse than a wattz.


chemistrymain2:
This is the stupidest PR i've ever seen.

PR #384, opened 2022-07-22
Vendor rebalancing

OliOliOnsiPree:
## About The Pull Request
So basically this PR nerfs and balances the vendors. So that way vendors can be actually balanced around the caps and actually can be used to the intended feature is without earning a thousand of caps from one vendor, at least this fixes the problems that was having for right now. Tweaks the metal price to 0.5, so that way people don't game it too much. 
Please give feedback if you dont agree with the changes. :) 

## Pre-Merge Checklist

## Changelog
:cl:
tweaks: vendor prices.
adds: Wildcard box in special vendor.
remove: musicbox that has been borked in years and proto thing from mining vendor. (You can make these)
/:cl:


Tk420634 (approved):
Lets give it a go!

PR #386, opened 2022-07-22
Escapes opening square brackets which were causing MDB to fail

Shadow-Quill:
## About The Pull Request
Byond considers [square brackets] in quotes to be replacement bits for a variable in the square bracket
ie: "There are [number_of_apples] in this pile.", where number_of_apples is a variable.

Which is why MDB is failing, because map files can't be changed during compile-time.

## Pre-Merge Checklist


Shadow-Quill:
I have been informed by the wonderful affectedmark in #coderbus that MDB is failing because it can't render the old version of the map.

After this PR is merged, any further renders of the map _should_ work fine.

image

Tk420634:
Can you boop me tomorrow sometime in the afternoon CST?  I'd really like to sit and make stupid cat noises at you.  Not just about this, but kind of in general.

Tk420634 (approved):
do eet

PR #392, opened 2022-07-23
Hospital udate 2 electric boogaloo

ZeroNetAlpha:
## About The Pull Request

Fixes the basement having walls in places it shouldn't be having walls as well as renames the hospital area to Christus Saint Michaels Hospital to fall in line with landmarks and the location of the area. 
## Pre-Merge Checklist

## Changelog
:cl:
tweak: changed the basement a bit because i left walls where i shouldn't have.
tweak: renames hospital area from Yuma General to Christus Saint Michaels to better fit with our lore.
/:cl:


Tk420634 (approved):
Thank you so much for changing the area name, my dad worked on this hospital when it was being built and it means a lot to me.  <3

PR #436, opened 2022-07-27
Smoke Signals

fludd12:
## About The Pull Request

A new communication technique for the tribals, and a minor change to bonfires.
- Bonfires can now be surrounded with three sandstone bricks to make them unable to be crossed, preventing you from lighting yourself on fire accidentally.
- If you speak tribal, using a sheet of cloth on an outdoor bonfire that does not have a grill will allow you to send a smoke signal message. All other people who speak tribal who are outdoors will be able to see the message, and if they're in the same Z-level, the rough direction it came from, but not who sent it.

## Pre-Merge Checklist

## Changelog
:cl: Fel
add: Those who know the guarded language of the local tribes can read and write smoke signals by carefully burning a sheet of cloth at a bonfire. Just make sure it isn't under a roof, or covered with a grill!
add: Tribals and villagers alike will appreciate the new and improved technique of bonfire construction, where adding just a few bricks of sandstone to the base will make it much harder to trip and fall in.
/:cl:


fludd12 (commented):
Daturana sutures are redundant

fludd12 (commented):
Whoops, missed these.

code/modules/hydroponics/grown/towercap.dm
```diff
@@ -1,3 +1,5 @@
+#define OUTDOOR_AREAS list(/area/f13/wasteland, /area/f13/desert, /area/f13/farm, /area/f13/forest, /area/f13/ruins)
```
Superlagg, line 1: I'd replace this with a GLOBAL_LIST_INIT(outdoor_areas, list()). I can see this list being useful for other things too!

code/modules/hydroponics/grown/towercap.dm
```diff
@@ -179,6 +182,48 @@
 	return ..()
 
 /obj/structure/bonfire/attackby(obj/item/W, mob/user, params)
+	//SMOKE SIGNALS. Burn cloth to send a radio message to everyone who knows tribal.
+	if(istype(W, /obj/item/stack/sheet/cloth) && !grill && burning)
+		var/obj/item/stack/sheet/cloth/C = W
```
Superlagg, line 187: I'd prefer this all to be its own proc, and use more descriptive vars.

code/modules/hydroponics/grown/towercap.dm
```diff
@@ -179,6 +182,48 @@
...
+				break
+		if(!outdoors)
+			to_chat(user, span_warning("You must be outside to send a smoke signal."))
+			return
+		var/signalmessage = stripped_input(user, "What would you like to send via smoke signal?", "Smoke Signal")
+		if(!signalmessage)
+			return
+		var/send_time = length(signalmessage) * 2 //Deciseconds. A 10 character message would take 20 deciseconds, or two seconds, to send.
+		if(!user.canUseTopic(src, BE_CLOSE))
+			to_chat(user, span_warning("You have to stand still to send a smoke signal."))
```
Superlagg, line 203: After we have a valid signalmessage, I'd prefer if the failure messages also included that message, so its not lost and players can just copypaste it back in.

code/modules/hydroponics/grown/towercap.dm
```diff
@@ -179,6 +182,48 @@
...
+				to_chat(user, span_warning("You don't have enough cloth to send a smoke signal."))
+				return
+			smoke_signal(user, signalmessage, src)
+			user.visible_message("[user] burns \the [W] in short, controlled bursts.")
+		else
+			to_chat(user, span_warning("You have to stand still to send a smoke signal."))
+		return
+	//SANDSTONE GUARDRAIL. Use sandstone to make the bonfire dense to avoid accidentally walking onto the fire.
+	if(istype(W, /obj/item/stack/sheet/mineral/sandstone) && !stones)
+		var/obj/item/stack/sheet/mineral/sandstone/S = W
```
Superlagg, line 216: I'd prefer a more descriptive var than S. Trying to get away from one-letter vars.

That said, thank you thank you thank you for this change~

Superlagg (changes requested):
Very good PR, very fitting for tribals and something Ive wanted for a long time! A few changes Id like to see though!

code/modules/hydroponics/grown/towercap.dm
fludd12: Will do!

code/modules/hydroponics/grown/towercap.dm
fludd12: That's a very good point. I wish there was a way to make it automatically fill in the last attempted message if it failed, but that would require setting a variable _somewhere_, and I don't think the player, the cloth, _or_ the bonfire really needs a variable like that.

code/modules/hydroponics/grown/towercap.dm
fludd12: The main reason it's a one-letter var is just because this is a typecast used to make sure the `use()` proc properly uses one sheet of sandstone. That being said, I can expand it!

code/modules/hydroponics/grown/towercap.dm
Superlagg: We already have the message stored in `signalmessage`, would be enough to just do something like:

`to_chat(user, span_warning("You don't have enough cloth to send a smoke signal. original message: [signalmessage]"))`

code/modules/hydroponics/grown/towercap.dm
Superlagg: I know, I just really dont like one-letter vars. Not that big a problem here tbh.

Superlagg (changes requested):
Almost there!

Superlagg (approved):
Looks good to me!

PR #439, opened 2022-07-27
sacrificing weapons and mobs on tribal altar

DrNukesky:
## About The Pull Request
now you can sacrifice weapons and mobs on the altar. When sacrificing mobs with a probability of 95%, they will drop random seeds, 5% chance of getting interesting potions. When sacrificing weapons, they will disintegrate into base materials, with a 5% chance of dropping rare components.

To sacrifice mob just buckle it to the altar and hit altar with bone talisman
To sacrifice weapon just hit altar with any gun. ONLY guns are acceptable, no melee weapons.

## Pre-Merge Checklist

## Changelog
:cl:
add: you can sacrifice weapons and mobs to the altar
/:cl:


code/game/objects/structures/divine.dm
```diff
@@ -1,3 +1,20 @@
+GLOBAL_LIST_INIT(potion_rates, list(
+	/obj/item/slimepotion/slime/sentience		=9,
```
Superlagg, line 2: These I'd be very wary about. Sentience potions, afaik, can be used on any fo13 mob, meaning possible sentient deathclaws (which, thinking about it, I kinda like). Speed potions totally negate armor slowdown, making salvaged armor basically better than PA. Slimeradio... I think it just works on mobs, but it gives them a built in radio. And Im not sure if slime peace can be reversed.

I'd take out the slime speed, and maybe the peace potion depending on how it worms.

code/game/objects/structures/divine.dm
```diff
@@ -1,3 +1,20 @@
...
+	/obj/item/slimepotion/speed		=2,
+	/obj/item/slimepotion/slime/renaming			=9,
+	/obj/item/slimepotion/slime/slimeradio		=9,
+	/obj/item/slimepotion/peacepotion		=9
+))
+
+GLOBAL_LIST_INIT(gunparts_rates, list(
```
Superlagg, line 9: I'd prefer if instead of a fixed list for all guns, something to make it based off the quality or rarity of the gun. Like, basic hoboguns would give at most a couple parts, while an assault rifle would give some serious parts, and a minigun or gauss rifle would basically explode into valuables.

Also, might want to check out the crafting recipes for various guns, and make sure it won't end up with a net positive. At least not a reliable one, if its chance based whether it'll actually happen, that should be fine.

OliOliOnsiPree:
You are going to hate me through your veins, but the potions might be a big no for right now.

code/game/objects/structures/divine.dm
```diff
@@ -6,19 +23,47 @@
...
-/obj/structure/sacrificealtar/on_attack_hand(mob/living/user, act_intent = user.a_intent, unarmed_attack_flags)
+/obj/structure/sacrificealtar/attackby(obj/item/I, mob/user)
 	. = ..()
+	var/divine = prob(divine_chance)
 	if(.)
 		return
+	if(istype(I, /obj/item/gun))
+		to_chat(user, "<span class='notice'>You attempt to sacrifice [I] by invoking the sacrificial ritual, it melts and breaks into pieces.</span>")
+		qdel(I)
+		message_admins("[ADMIN_LOOKUPFLW(user)] has sacrificed [key_name_admin(I)] on the sacrificial altar at [AREACOORD(src)].")
```
Superlagg, line 37: I'd prefer this changed to a log, admemes get a lot of messages on average, and we're not really concerned about people deleting a dozen hipowers and knuckleguns, unless someone complains.

code/game/objects/structures/divine.dm
```diff
@@ -6,19 +23,47 @@
...
 	if(!has_buckled_mobs())
 		return
 	var/mob/living/L = locate() in buckled_mobs
 	if(!L)
 		return
-	to_chat(user, "<span class='notice'>You attempt to sacrifice [L] by invoking the sacrificial ritual.</span>")
-	L.gib()
-	message_admins("[ADMIN_LOOKUPFLW(user)] has sacrificed [key_name_admin(L)] on the sacrificial altar at [AREACOORD(src)].")
+	if(istype(I, /obj/item/clothing/accessory/talisman))
+		to_chat(user, "<span class='notice'>You attempt to sacrifice [L] by invoking the sacrificial ritual, during the sacrifice, you removed something from the stomach of the victim.</span>")
```
Superlagg, line 53: I'd prefer if you had to kill the mob first. Or just have that be a requirement for non-simple mobs. 

I also dont like the idea of straight up gibbing player mobs, I'd prefer something like marking them as sacrificed so they cant be sacrificed again, or destroying their brain to keep them dead for good.

code/game/objects/structures/divine.dm
```diff
@@ -6,19 +23,47 @@
...
+		L.gib()
+		message_admins("[ADMIN_LOOKUPFLW(user)] has sacrificed [key_name_admin(L)] on the sacrificial altar at [AREACOORD(src)].")
+		switch(divine)
```
Superlagg, line 56: I'd also like some kind of tier system to mobs, so a chicken isnt worth as much as a deathclaw. Plus, it could mean people spamming easy to make mobs for lots of chances at potions. Not really concerned about all the seeds though, maybe low-level mobs would just check for seeds?

Superlagg (changes requested):
Very good PR, something we've been needing! A new things to consider though.

code/game/objects/structures/divine.dm
DrNukesky: I don't see much point in this yet, since deathclaws and other large and dangerous mobs currently cannot be moved due to their size, which means there is no way to put them on the altar. The rest of the mobs are no different in their danger. To an equally mild degree, brahmins, chickens, or mirelurks, who have a habit of forming whole packs in front of the camp, can be worn on the altar. I will gladly try to add such a tier list when hunting for mobs becomes a real challenge, and not a routine task for 2 minutes

code/game/objects/structures/divine.dm
DrNukesky: Removed those potions

code/game/objects/structures/divine.dm
DrNukesky: The basic idea of ​​sacrificing weapons is that they melt right on the altar, and you have to be very lucky to extract at least something useful from the remains. In any case, the loot obtained in this way does not exceed the cost of crafting any weapon, so this system will not be abused.

code/game/objects/structures/divine.dm
DrNukesky: Fair point. These messages are left over from the original version of the file. I think it's best to remove them altogether.

code/game/objects/structures/divine.dm
DrNukesky: Good idea, but I don't know how to do it. If someone tells me the code to remove the brain of the victim, I will gladly write it.

but I added a check that the victim must be dead.

code/game/objects/structures/divine.dm
Superlagg: Fair enough, we'll see how it works out and make changes later if needed.

code/game/objects/structures/divine.dm
Superlagg: Fair enough!

code/game/objects/structures/divine.dm
Superlagg: Should be okay for now, I see it mostly just for livestock, with the very occasional player being sacrificed for RP reasons.

code/game/objects/structures/divine.dm
Superlagg: thanks~

Superlagg (approved):

Superlagg (approved):

code/game/objects/structures/divine.dm
```diff
@@ -6,19 +21,50 @@
...
+	var/divine_chance = 5
 
-/obj/structure/sacrificealtar/on_attack_hand(mob/living/user, act_intent = user.a_intent, unarmed_attack_flags)
+/obj/structure/sacrificealtar/attackby(obj/item/I, mob/user)
 	. = ..()
+	var/divine = prob(divine_chance)
 	if(.)
 		return
+	if(istype(I, /obj/item/gun))
+		to_chat(user, "<span class='notice'>You attempt to sacrifice [I] by invoking the sacrificial ritual, it melts and breaks into pieces.</span>")
```
Shadow-Quill, line 33:
```suggestion
		to_chat(user, span_notice("You attempt to sacrifice [I] by invoking the sacrificial ritual, it melts and breaks into pieces."))
```

code/game/objects/structures/divine.dm
```diff
@@ -6,19 +21,50 @@
...
 	if(!L)
 		return
-	to_chat(user, "<span class='notice'>You attempt to sacrifice [L] by invoking the sacrificial ritual.</span>")
-	L.gib()
-	message_admins("[ADMIN_LOOKUPFLW(user)] has sacrificed [key_name_admin(L)] on the sacrificial altar at [AREACOORD(src)].")
+	if(L.stat != DEAD)
+		to_chat(user, "<span class='notice'>You should kill it first!.</span>")
+		return
+	if(istype(I, /obj/item/clothing/accessory/talisman))
+		to_chat(user, "<span class='notice'>You attempt to sacrifice [L] by invoking the sacrificial ritual, during the sacrifice, you removed something from the stomach of the victim.</span>")
```
Shadow-Quill, line 54:
```suggestion
		to_chat(user, span_notice("You attempt to sacrifice [L] by invoking the sacrificial ritual, during the sacrifice, you removed something from the stomach of the victim."))
```

Shadow-Quill (changes requested):

Superlagg (approved):

Issue #484, opened 2022-08-01
Skills on spawn


Shadow-Quill:
Spawning with certain skills, like the overseer spawning with chem knowledge, doesnt give them the recipes associated with that trait's book, and prevents you from reading that book.

So basically, have them spawn with the skill book, instead of the skill.

Potentially related to #425

Tk420634:
Pretty much murdered in the massive skill overhaul.

Issue #425, opened 2022-07-26
Chem Wizard perk from character creation doesn't grant drug recipies


Jamini:
---- Testmerges: ----
Live Code

---- Reproduction: ----

1. Select the Chem Wizard Trait from the perk selection
2. Spawn
3. Open Medical Tab
4. Jet and Turbo recipies are absent

---- Additional context: ----

All other forms of getting the chem wizard trait (including spawning as a follower) allow you to make Jet and Turbo

PR #521, opened 2022-08-05
Gives some quirks their recipes

Ghostcoffeee:
## About The Pull Request

I've seen some complaints that the chem-whiz quirk doesn't give the recipes you'd get if you read the book. This fixes that, as well as the explosive crafting quirks

I don't think any other quirks should have recipes, but if they do, complain at me

Should fix issues #484 and #425

## Pre-Merge Checklist

## Changelog
:cl:
add: Gave recipes to the chem-whiz and explosives crafting quirks
/:cl:

### DIFF
--- a/code/datums/traits/good.dm
+++ b/code/datums/traits/good.dm
@@ -1,3 +1,35 @@
+GLOBAL_LIST_INIT(chemwhiz_recipes, list(
+    /datum/crafting_recipe/jet, 
+    /datum/crafting_recipe/turbo, 
+    /datum/crafting_recipe/psycho, 
+    /datum/crafting_recipe/medx, 
+    /datum/crafting_recipe/medx/chemistry, 
+    /datum/crafting_recipe/stimpak, 
+    /datum/crafting_recipe/stimpak/chemistry, 
+    /datum/crafting_recipe/stimpak5, 
+    /datum/crafting_recipe/stimpak5/chemistry, 
+    /datum/crafting_recipe/superstimpak, 
+    /datum/crafting_recipe/superstimpak5, 
+    /datum/crafting_recipe/buffout,
+    /datum/crafting_recipe/steady))
+
+GLOBAL_LIST_INIT(basic_explosive_recipes, list(
+	/datum/crafting_recipe/frag_shrapnel, 
+	/datum/crafting_recipe/stinger, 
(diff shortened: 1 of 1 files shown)


code/datums/traits/good.dm
```diff
@@ -327,6 +327,43 @@
 	lose_text = "<span class='danger'>You forget how the periodic table works.</span>"
 	locked =  FALSE
 
+/datum/quirk/chemwhiz/add()
+	var/mob/living/carbon/human/H = quirk_holder
+	// I made the quirks add the same recipes as the trait books. Feel free to nerf this
+	H.mind.learned_recipes += list(
```
Superlagg, line 333: Please turn this list and the others into global listw, that way they can be used for the other dozen or so other things that use literally the same list. We're trying to refactor these skills into global lists.

code/datums/traits/good.dm
```diff
@@ -426,6 +485,41 @@
 	lose_text = "<span class='danger'You feel like you're allowed to fly on planes again.</span>"
 	locked = TRUE
 
+/datum/quirk/advanced_explosive_crafting/add()
+	var/mob/living/carbon/human/H = quirk_holder
+	// I made the quirks add the same recipes as the trait books. Feel free to nerf this
+	H.mind.learned_recipes += list(
+		/datum/crafting_recipe/incendiary,
```
Superlagg, line 492: This can be broken up into two global lists, one for the basic explosive crafting recipes, and one for what the advanced explosive crafting adds to that, and have the two lists added just like this.

code/datums/traits/good.dm
```diff
@@ -417,6 +454,28 @@
 	lose_text = "<span class='danger'You feel okay with the advancement of technology.</span>"
 	locked = FALSE
 
+/datum/quirk/explosive_crafting/add()
+	var/mob/living/carbon/human/H = quirk_holder
+	// I made the quirks add the same recipes as the trait books. Feel free to nerf this
+	H.mind.learned_recipes += list(
```
Superlagg, line 460: This should also be a global list.

Superlagg (changes requested):
Very good so far, something thats needed to be done for a while! A few things I need changed though.

Superlagg (approved):
looks good!

PR #555, opened 2022-08-12
Gun changes, again

Superlagg:
## About The Pull Request

Manually cycled guns can be racked by just clicking again, so you can just keep clicking to rack your gun.

Pumping the gun is no longer tied to the gun's firing rate, so you can pump/rack it immediately after firing. However, the firing rate itself still limits how often the gun can be fired. 

Messed with the ranged click delay so that you'll cock your gun quickly after firing, instead of just before firing. Just some feely things.

Fun fact, the rifleman trait didnt really do much in the new system! Now it does, it reduces the refire rate on manual rifles and shotguns by 20%. I know its a lot less than 50% faster, but, now it actually works.

Certain guns have different chances to hit specific zones on people. Most semiauto guns work as they did, but automatic guns have a much lower chance to hit your chest or head beyond a few tiles, and shotguns heavily favor hitting limbs if not point blank. However, manual / precision guns will always hit your selected zone, unless its the head, which requires you to be within 7 tiles for a guaranteed hit, otherwise its just a 20ish% chance to hit the head (and like 60% to hit the torso instead).

Complicated, but tldr, less precise guns fuck up limbs more, precision guns fuck up what you aim for.

Also rubber ammo typically forces precision limb aiming, if it isnt shot through an automatic.

Also added in support for damage falloff. Its currently unused, but if yall dont behave, I'm turning this PR right around!

## Pre-Merge Checklist

## Changelog
:cl:
add: Manual guns are racked by clicking too.
add: Rifleman skill actually works now, reduces fire delay on manual guns by 20%
tweak: Imprecise guns aim for limbs more, precise guns aim for what you picked more.
/:cl:

### DIFF
--- a/code/__DEFINES/combat.dm
+++ b/code/__DEFINES/combat.dm
@@ -311,6 +311,41 @@ GLOBAL_LIST_INIT(main_body_parts, list(
 #define BULLET_DAMAGE_SHOTGUN_PELLET 11
 #define BULLET_DAMAGE_SHOTGUN_SLUG 50
 
+/// Bullet damage falloff per tile defines
+#define BULLET_FALLOFF "bullet falloff per tile"
+#define BULLET_FALLOFF_PISTOL_LIGHT 3
+#define BULLET_FALLOFF_PISTOL_MEDIUM 5
+#define BULLET_FALLOFF_PISTOL_HEAVY 6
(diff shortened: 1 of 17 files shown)


Shadow-Quill (commented):
Code looks good from a quick scroll through, but I abstain from commenting on the balance prods.

thingpony (approved):

PR #564, opened 2022-08-13
Reduces town doctor coun t to 1

Superlagg:
## About The Pull Request

Town gets one doc,


thingpony (approved):
One doctor to rule them all, one doctor to find them, One doctor to bring them all, and in the darkness bind them; In the Land of Coyote where the bayous lie.

PR #614, opened 2022-08-20
Removes the BYOND advertisement

Marrone321:
Sorry Lummox...

Removes the "Become a BYOND member to access member-perks and features, as well as support the engine that makes this game possible. Only 10 bucks for 3 months!" message

There are other things connected to the proc that have to do with BYOND membership and admin stuff, so I decided to let the proc stay.


Superlagg:
I dont like the idea of removing the advertisement, but if you believe its the right thing to do, I'll approve it.

Marrone321:
Requested by Nayriin#1860

Superlagg (approved):
shrug

PR #642, opened 2022-08-25
You can stick your survival kit back in your pocket

Superlagg:
## About The Pull Request

Whatever, put the kit back in your pocket.

Increased storage spaces in most belts and neck-things by 1 slot.

Also:

Bio and chem bags use the same storage allowed stuff as medibelts, so you can put your whole kit in there.

Assault and army belts are bandoliers now.


Superlagg:
got something better coming!

Issue #631, opened 2022-08-23
Light Tribal Cloak hides clothing and shows underwear.


Brimcon:
As said on the tin, the Light tribal cloak hides any uniform/clothing worn and shows underwear/undershirt if there is any. Best seen on full body uniforms such as settler's outfits or any contrasting color clothes.

Superlagg:
fix't

PR #654, opened 2022-08-27
Tribal Cloak no be nakey

thingpony:
Fixes the tribal cloak/armor hiding your jumpsuit and force viewing your undies/gentials when you don't want it too

Also fixes some bad paths in ears


Brimcon:
Just mentioning this closes #631

Superlagg (approved):

PR #655, opened 2022-08-27
Fixes survival kits being invisible

Superlagg:
## About The Pull Request

So turns out soome of them were invisible. Whoops.


Superlagg:
uh i dont really know what i did and if anything breaks its probably cus of this

PR #660, opened 2022-08-28
Virgo-style storage UI

Superlagg:
## About The Pull Request

Switches over to a bay/polaris/virgo style storage UI based on volume and such over some kind of nebulous slot n size thingy. Hated that, and I like this, so we're using it.

Also turns a bunch of jank component initialization storage shit into actual storage datums, so not everything is some bespoke snowflake or something. Easier to change values and such, for sure.

Also adds one more slot to all survival pouches.

Also makes combat armor into a bandoler/holster.

Duffelbags slow you a bit less and hold a bit more.

Torches are smol.

Most cool hats have a storage slot.

Chem / bio / medi bags have the same allowed stuff.

Adds an option for volumetric storage to respect max items (and look good too!). This is on by default for all non-numeric storage, and off by default on backpacks.


Shadow-Quill (approved):
quick scroll-through of the code doesn't show anything that sticks out.
definitely should be _test_merged, and also prolly should add a few screenshots for examples

Tk420634 (approved):

PR #757, opened 2022-09-11
Adds in Kuba's gear!

Superlagg:
## About The Pull Request

Gives a yeen a sledgehammer and some downtiered armor.


modular_citadel/code/modules/client/loadout/__donator.dm
```diff
@@ -137,3 +137,12 @@
 	new /obj/item/gun/energy/laser/wattz(src)
 	new /obj/item/stock_parts/cell/ammo/ec(src)
 	new /obj/item/stock_parts/cell/ammo/ec(src)
+
+/datum/gear/donator/kits/kuba
+	name = "Earlong Travel Supplies"
+	path = /obj/item/storage/box/large/custom_kit/kuba
+	ckeywhitelist = list("pappavol")
+
+/obj/item/storage/box/large/custom_kit/kuba/PopulateContents()
```
Shadow-Quill, line 146: ideally, we should just be using the ckey in the path

Superlagg: Should be fixed

Tk420634 (approved):

PR #785, opened 2022-09-14
ci skip'd commits will auto-pass the linter check

Shadow-Quill:
As per title, commits with `[ci skip]` in the message will skip CI as usual, but will now automatically pass the `Run Linters` step, as it's a required step and otherwise won't actually ever return a status.

No player-facing changes.


Shadow-Quill:
welp, this won't work. thanks github :D

PR #812, opened 2022-09-19
changes attack speed and damage of melee weapons

bearrrrrrrr:
## About The Pull Request
as the above says. way too many changes to changelog.

## Pre-Merge Checklist

## Changelog
:cl:
tweak: some melee weapons attack faster or slower
balance: melee weapons are overall better, generally
/:cl:

### DIFF
--- a/code/game/objects/items/melee/f13onehanded.dm
+++ b/code/game/objects/items/melee/f13onehanded.dm
@@ -155,6 +155,7 @@
 	bare_wound_bonus = 5
 	custom_materials = list(/datum/material/iron=6000)
 	resistance_flags = FIRE_PROOF
+	attack_speed = CLICK_CD_MELEE * 0.85 //6.8
 
 /obj/item/melee/onehanded/knife/Initialize()
 	. = ..()
@@ -181,7 +182,6 @@
 	force = 23
 	throwforce = 25
 	attack_verb = list("slashed", "stabbed", "sliced", "torn", "ripped", "cut")
-	attack_speed = 6.8
 
 /obj/item/melee/onehanded/knife/survival
 	name = "survival knife"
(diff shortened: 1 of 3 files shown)


out-of-phaze:
Why use `= 6` and `= 6.8` etc. rather than `CLICK_CD_MELEE * 0.8`, `CLICK_CD_MELEE * 0.85`, etc.? Isn't the latter much easier to understand? They're the same values regardless.

bearrrrrrrr:
> Why use `= 6` and `= 6.8` etc. rather than `CLICK_CD_MELEE * 0.8`, `CLICK_CD_MELEE * 0.85`, etc.? Isn't the latter much easier to understand? They're the same values regardless.

Because I was a sleepy bitch. Fixing rn

code/game/objects/items/melee/f13onehanded.dm
```diff
@@ -155,6 +155,7 @@
 	bare_wound_bonus = 5
 	custom_materials = list(/datum/material/iron=6000)
 	resistance_flags = FIRE_PROOF
+	attack_speed = CLICK_CD_MELEE * 0.85 //6.8
```
Superlagg, line 158: Id prefer the 0.85 and such to also be defines, so they're easier to change later if needed. You can put them in combat.dm's defines!

Superlagg (changes requested):

code/game/objects/items/melee/f13onehanded.dm
Superlagg: Same for the armor penetration values.

bearrrrrrrr: The reason I didn't make them defines was because a lot of them have smaller tweaks. It feels like it'd be redundant and just weird. I will do it if you _really_ want to, but I don't think it's needed.

Superlagg: Should be fine for now then!

Superlagg (approved):

PR #837, opened 2022-09-25
Tool mod adjustments/Addition! + ckey kit

AmethystTheSergal:
Loadout kit locked to ckey Gozulio.

Adds in the Vault-Tec Integrated ZAX module crafting recipe. Requires guns and bullets three plus an advanced workbench. Also changes that mod up somewhat.

Drastically lowers the chargecost increase of the battery shunt. NO WONDER PEOPLE DIDNT USE IT BEFORE. The booster is superior with lower cost increase and just 10% less damage. Why on earth would you eat a 3 times chargecost increase for just 10% damage increase? *angry wolf sounds*


AmethystTheSergal:
For reference: The booster charge cost is 35 increase. While the battery shunt is 200%. mmmh yes I love 3 times the cost to my poor battery for just a 10% boost. MMM. Battery shunt more like Battery melter.

Tk420634 (approved):

PR #840, opened 2022-09-25
I totally know balance (Adjustment wild ride)

AmethystTheSergal:
Changes up various mods numbers (see the code tab).

This should hopefully fix up things to not be too absurd.


AmethystTheSergal:
From Dev chatter:

overdrive lowers damage by 20% more
(50% reduction)
giga lens had their chargecost be more in line with the other two damage increasers 
raider trigger fire delay has been lowered by 5%
(15% increase)
charge cost reduction of the AI tool lowered to 10% down from 20%

Superlagg (approved):

PR #843, opened 2022-09-25
Bloodloss and bleeding overhaul, take II

Superlagg:
## About The Pull Request

Heavy reworks to how bloodloss fucks you and refactors to bleeding wounds.

 - Increases total blood volume from 560ish to 1000.
 -  - Reason: More wiggle room for a larger range of blood-level based effects, allows people to bleed more without being completely disabled, and slow down recovery.
 - Blood loss kills at 0%, instead of 100%
 -  - Reason: The only functional difference between 50% and 0% before was how much blood you'd put in before reviving them. Killed and perma stunned players too quickly.
 - Blood loss causes various different debuffs depending on your level of blood loss, such as capping your oxy/stamina, blurring, missteps, sprint debuffs, etc.
 -  - Reason: The only effect bloodloss had before was making you go from just fine to perma stunned after some mild blood loss. This way, players can suffer through the whole range of blood loss while conscious.
 - Each level of blood loss will have its own soft cap of how low it can take your blood level. Such as, small scratches wont bleed you further than minor symptoms, you'd need a weeping avulsion to have any chance of getting turbofucked
 -  - Reason: Prevent trash mobs from totally tanking your blood level, while still making them a threat if you're gonna be adventuring anywhere more difficult.
 - Most mobs can make you bleed. This includes ghouls, handies, radscorpions, and just about anything else that's reasonably sharp enough.
 -  - Reason: Makes blood loss a lot more prominent.
 - Increased wound defence on most armors. Goal is to have light armor protect you from low-trash mobs like geckos, medium from ghouls and light bullets, and heavy from anything less than a hunting rifle.
 -  - Reason: Require people to put some armor on before going into a combat situation, or git good. Also make medium+ armor more useful.
 - Most things that used to replenish blood, now instead act as temporary blood while in your system. This includes things like water, xander tea, saline, bloody mary, and most consumables.
 -  - Reason: Allow for temporary solutions to blood loss, without sidestepping the actual problem and requiring people actually treat their blood loss.
 - Epinephrine prevents your effective blood volume from going below just under the tier for the worst effects, and halves your blood loss.
 -  - Reason: Make epinephrine more useful to have on hand, so people dont just ditch the darn thing.
 - Sanguirite reduces your blood loss by around 95% while in your system.
 -  - Reason: Make them useful for blood loss, without negating the need to actually treat your wound.
 - Nutrition helps you regenerate your blood a lot faster than before. The exchange between blood regenerated and nutrition spent doesn't change, but if you're more well fed, you'll 'buy' more blood per life cycle.
 -  - Reason: Give more of a role to food, and encourage people to bring along a snack while adventuring.
 - Oxy loss knocks you out at 90 damage instead of 50.
 -  - Reason: More wiggle room for blood loss oxyloss effects.
 - New damage type: Wound damage!
 - - As limbs take damage from sharp damage sources, they accumulate wounding damage. If a sharp attack raises it past a wound threshold, or you take wounding damage while above that threshold, you take a wound based on how much wound damage is on the limb.
 - - Drastically reduced the amount of RNG behind wounding, now its all based on after-armor damage + wound modifier.
 - - Reason: The old system was impossible to balance around and didnt work properly. Now? It does.

Major changes to bandages and sutures!

 - Bandages and sutures generally need just one application of each for their full effect.
 -  - Reason: more intuitive to just stick one on instead of just stuffing them into a wound till they're treated.
 - Bandages and sutures take damage if you take damage.
 -  - Reason: Encourage people to slow down and hang back while they heal, or just stop getting hit.
 - Bandages and sutures apply their full effect over their entire duration, which is typically between 20 minutes and an hour based on tier.
 -  - Reason: Much simpler for me to code.
 - Bandages and sutures (mostly sutures) get a bonus to their speed to use and effectiveness if either you or your target has some medical training (traits).
 -  - Reason: Encourages people to bring along a doctor, so you can't just self-medicate as easily.
 - Bandages slow bleeding by 95ish% while on their limb.
 -  - Reason: Make them useful without stepping on sutures' role.
 - Bandages heal both brute and burn.
 -  - Reason: Help players stay healed, so blood loss is their main concern. Also sutures generally don't heal you.
 - Bandages are quick to apply.
 -  - Reason: Make them useful without stepping on sutures' role.
 - Sutures generally hurt you when used. They are also slow to apply, unless you're a doctor.
 -  - Reason: It's a needle and thread.
 - Sutures completely stop your bleeding on that limb while on.
 -  - Reason: Make them useful without stepping on bandages' role.
 - Sutures heal the wound over time, quicker if the applyer is a doctor, if the sutures are higher tier, you are well fed, and/or there are bandages on that limb.
 -  - Reason: Make them useful without stepping on bandages' role.
 - Bloodpacks hold 1000 units. They also can't be directly filled without using an IV, but can be syringed out of.
 -  - Reason: People have a LOT more blood in them, and are gonna need more blood.
 - Bloodbags (and radaway cus its a child of bloodbags) need a syringe or IV pole to get anything out of
 -  - Reason: Yeah just squirt that blood in your mouth, that'll treat your massive bleeding.

Changes to medscanners!

 - More obvious, accurate readouts of bleeding wounds, complete with which limb, the name of the wound, the wound size, and the total and effective bleed rate.
 -  - Reason: Surprise this wasn't part of them to begin with. honestly.


bearrrrrrrr:
Finally have a good way to format my worries: my beeeeg concern here is ending up at a point where people will be normally walking around missing a good chunk of blood or having an amount of oxydamage.

Besides that: very excited to see this rework. Excited to see how and where it goes.

Tk420634 (approved):

PR #845, opened 2022-09-26
*Uses a nerf bat on harm intent* (overdrive chip addition)

AmethystTheSergal:
Charge cost lowering is now 60% of the overdrive, but it lowers damage by 75%. You want full auto? You get full auto. But those will tickle without a damage upgrade here and there.

All damage increase numbers of the energy weapon upgrades are lowered by 5% each.


AmethystTheSergal:
Reason for nerf:

It has full auto. Full auto is very good on its own.

Tk420634 (approved):

PR #849, opened 2022-09-26
Turret overhaul

Superlagg:
## About The Pull Request

Overhauls turrets!

Turret AI has been completely changed.

When idle, it scans for targets as normal.

Once it finds a target, it hyperfocuses on that for the next several seconds, spraying them, and only them, with bullets.

If the target is no longer visible, it shoots at where they were. They won't try to look for the actual target until it fully loses interest.

All the while, it'll shine a laser pointer at its target. Doesnt do anything other than look cool and show what its shooting at.

In this time while the turret has you or your turf locked on, someone else could just run up to the turret and destroy it. Strategy!

If the turret has been firing at nothing for a while, eventually it will get bored and go back to scanning. It'll make some cute beeping noises when it sends out a sensor ping. If you time it right, you can run past them in between pings.

Turrets now make noises when they do things, such as:
 - Being alerted
 - Seeing something they dont like
 - Looking for something
 - Going to sleep

Turrets have a variable bullet spread, so they're no longer perfectly accurate.

Turrets also can bire bursts of shots. Dont worry, it checks before every shot if its target is awake, and will immediately stop if they go down.

Turret gunfire noises can be heard from a distance, like guns.

Turrets no longer use power, currently. Not that there's finite power, but it saves some checks.

The GUI is busted right now, but it'll be able to pick some useful targets now.


bearrrrrrrr:
this is so fucking sick omfg. do the turrets change targets if they get shot at by them?

BUT THIS IS SO SICK THOUGH

Superlagg:
> this is so fucking sick omfg. do the turrets change targets if they get shot at by them?
> 
> BUT THIS IS SO SICK THOUGH

They have a chance to!

Superlagg:
So this'll be a draft until the maps are changed so that turrets are placed in a way that wouldnt make them utterly dominating, cus right now most turrets right now would become near-unstoppable furrymunchers, and not in the hot way.

Tk420634 (approved):

PR #856, opened 2022-09-28
Mob rebalancing

Superlagg:
## About The Pull Request

Mobs have armor now! Most of them just have light armor, but super mutants and sentrybots have medium+ armor. Lemme know how they are!

Adds in a description system to convey *more or less* how much armor the mob has.

There are two kinds of ranged supermutant now: Shotgun and Varmint Rifle ones.

Shotgun mutants shoot two improvised shells that will murder you... unless you're armored or more than like 4 tiles away. Then they'll just hurt a bit.

Varmint rifle mutants shoot a varmint rifle at you. Much weaker than a hunting rifle and less likely to wound you. Also much less likely to kill you.

Rebalanced mob health and armor for those who are armored.

EMP does various things to robots, not just damage them. Its possible to stunlock them like this!

Tidied up some simplemob files.

Also makes Fuzzy's rifle shoot slower. I dunno, they asked for it. Dont look at me like that!

### DIFF
--- a/code/__DEFINES/armor.dm
+++ b/code/__DEFINES/armor.dm
@@ -475,3 +475,53 @@ GLOBAL_LIST_INIT(armor_token_operation_legend, list(
 
 /// Damages DT is effective with
 GLOBAL_LIST_INIT(damage_threshold_valid_types, list("brute", "fire"))
+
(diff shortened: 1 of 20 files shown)


Superlagg:
k should be ready

Tk420634 (approved):

PR #922, opened 2022-10-07
Fixes potentially admin-tool breaking code

Yawet330:
## About The Pull Request

This replaces the system we were currently using to gatekeep the BoS with the proper code.

As it states. Please NEVER remove a job from its respective job-list unless it is the last and only solution, as if an admin decides to re-enable it midround, they will NOT show up beneath the faction, and other issues (like access to faction doors and/or gear) may occur.

For clarity sake, this only adjusts initiate slots up to 3, from 2 and that's because 4 people is _not nearly enough_ (that's if we have 1 scribe, 2 inits and 1 head-knight).

## Pre-Merge Checklist

## Changelog
:cl:
fix: Fixes some broken code that would otherwise cause a LOT of problems if we added systems to the BoS that began to rely on their job/faction list.
tweak: Bumps initiate slots by 1.
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/bos.dm
+++ b/code/modules/jobs/job_types/bos.dm
@@ -117,8 +117,8 @@ Head Paladin
 	title = "Head Paladin"
 	flag = F13SENTINEL
 	head_announce = list("Security")
-	total_positions = 1
-	spawn_positions = 1
+	total_positions = 0
+	spawn_positions = 0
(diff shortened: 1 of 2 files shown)


bearrrrrrrr:
Okay,

1. _**The reason why these jobs still pseudo-exist is because they are soon going to be re-added.**_ I have said this roughly five times, by this point.
2. Absolutely nobody needs that passive-aggressive code comment.

Glad to help.

Superlagg:
Did you test to see if this actually breaks things?

Tk420634 (approved):

PR #923, opened 2022-10-07
Stealthboys + Stealthsuit fix! (Needs host discussion)

AmethystTheSergal:
Adds stealthboys to the trash pile. 
Information on the stealthboys:
- Last a little bit before steadily revealing you.
- Doesn't work on simple mobs.
-  Provides no benefit anymore if used for prolonged periods. (basically turns into trash)

Probably fixed the stealthsuits sprite?


Tk420634:
> > > This seems really bad for our already sketchy PVP balance (which is non-existent), but will buff issues such as people stealing shit and INSTANTLY sprinting away
> > 
> > 
> > The problem is folks don't really have a desire to actually pvp because of carrying over 'karma'/remembering their character. With your response of 'instantly sprinting away', you've already found out why this is being tested!
> > If all goes ~horribly wrong cause this is a bad idea~ according to plan, it'll be changed up to be general loot in dungeons. (or you know when I wake up and decide to put it in the map as loot instead).
> 
> Entirely fair, my main concern was that since this has no real PVE gains, the only real reason to use it is for PVP/Other player versus player interactions (e.g thievery/stalking/infiltrating places). Which is the main core of the problem IMO. Even on a server that tends to avoid PVP...

God damn I hope they do.  The playerbases choice to avoid pvp is largely because how easy it is to get caught doing even mundane bad guy shenanigans.  Though I still firmly want certain mobs to be player playable instead, as they seem far more keen on playing short term unnamed baddies vs long term low karma characters.  Still feels like the option is good to be there, sides.  It's very easy to turn stuff back off.

As for the discussion part I'll be in voice tonight as always and we can fool around with this in a round with some semi-controlled environments.

AmethystTheSergal:
> God damn I hope they do. The playerbases choice to avoid pvp is largely because how easy it is to get caught doing even mundane bad guy shenanigans. Though I still firmly want certain mobs to be player playable instead, as they seem far more keen on playing short term unnamed baddies vs long term low karma characters. Still feels like the option is good to be there, sides. It's very easy to turn stuff back off.

As for the discussion part I'll be in voice tonight as always and we can fool around with this in a round with some semi-controlled environments.

Roger that! I'm available whenever with a ping. I'll also look over the code to see if it's possible to lower the duration of invisibility a bit.

AmethystTheSergal:
temporarily closing this pr! will reopen it after changes are fixed up elsewhere!

PR #1058, opened 2022-10-30
Fixes weather not working, like, at all

Superlagg:
## About The Pull Request

Weather is now shared across all Z levels! No more wierdness where one map is having a sunny day and another is a raging shitstorm!

Weather is also a lot more configurable as to where it happens. Areas have a list of weather types that are allowed to affect them, so you can have some areas that are affected by just rain, some by just radstorms, or even have some weather effects even affect you inside areas that're otherwise protected! No more fuckery with area subtypes!

Weather warnings should be broadcast to all above-ground non-admeme Z levels. Now if a radstorm is coming, and you're not in the sewers, you'll know! No more surprise storms!

Also fixes radstorms doing fuck all. Seriously, it was checking rad resistance twice, and then delivering a tiny dose even before armor! I changed it to work properly, and still deal a tiny bit of radiation, about as much as before, but its actually *intended* now!

If a nasty weather is happening on the other end of a ladder, it'll warn you and pop up a little rotary menu to confirm that yes, you want to run out into the fuckscape out there.

### DIFF
--- a/code/__DEFINES/weather.dm
+++ b/code/__DEFINES/weather.dm
@@ -1,0 +1,37 @@
+/// Z levels that are more or less above ground and can see the sky
+/// For telling players about the weather
+#define ABOVE_GROUND_Z_LEVELS list(4, 5, 6, 7, 8)
+
+/// Minimum time between weathers
+#define WEATHER_WAIT_MIN 30 MINUTES
+/// Maximum time between weathers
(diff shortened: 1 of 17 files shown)


Superlagg:
If you're underground and go to use a ladder, and the weather up there is frightful, you'll need to click on the ladder again to actually use the ladder.

Tk420634 (approved):

PR #1342, opened 2022-12-22
Makes armor tokens additive instead of multiplicative

Superlagg:
## About The Pull Request

Subclasses of armor should be a lot less ultra-volatile in how much they stray from their base values. Used to be that medium+ armor would have massive variance on their armor values. Now, they should more or less be within their armor rating, with less (but still some) difference between armor classes.

Kind of a testing thing!


PrettyPsychoForAWolf (approved):
Superlagg over here getting the good changes in. Thanks cutie.~

PR #1432, opened 2023-01-08
Pied Piper of Nash

NopemanMcHalt:
## About The Pull Request

Adds "Beastmaster - Rats" quirk. This allows the quirk holder to summon and control new type of rat vs other mobs.
They shouldn't attack quirk holder or other players unless they (or rats) got their faction changed (somehow). Also shouldn't attack dogs.
Curious mice also get changes to faction and *finally receive an ability to pet them*, not poke them.
Please note that it was done in rather hacky way, so quirk holder also gets ability to... Dig up nests for rats to appear.

## Pre-Merge Checklist

## Changelog
:cl: Anonymous
add: Rat Master(s) have arrived to Nash!
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/abilities_common.dm
+++ b/code/modules/mob/living/simple_animal/hostile/abilities_common.dm
@@ -10,6 +10,8 @@
 #define RTS_RATS_ALLOWED list(\
 		/mob/living/simple_animal/hostile/rat,\
 		/mob/living/simple_animal/hostile/rat/skitter)
+#define RTS_FRATS_ALLOWED list(\
+		/mob/living/simple_animal/hostile/rat/tame)
 #define RTS_ROBOT_ALLOWED list(\
 		/mob/living/simple_animal/hostile/handy,\
 		/mob/living/simple_animal/hostile/handy/protectron,\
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/simple_animal/hostile/abilities_common.dm
```diff
@@ -105,6 +107,9 @@
 	immune_to_lowpop = TRUE
 	allowed_mobs = RTS_RATS_ALLOWED
 
+/obj/effect/proc_holder/mob_common/summon_backup/rat/tame
+	allowed_mobs = RTS_FRATS_ALLOWED
+
 /obj/effect/proc_holder/mob_common/summon_backup/robot
 	allowed_mobs = RTS_ROBOT_ALLOWED
 	banned_from_lowpop = TRUE
```
Superlagg, line 115: Think I found why Summon Backup didnt work with humans, its the check on 118, and the var on 122. If you change if(!istype(user, /mob/living/simple_animal)) to if(!istype(user, /mob)) and var/mob/living/simple_animal/owner = user to var/mob/owner = user that should make it work on humans. =3

Superlagg (commented):
Good PR! Looks solid, does a good job with what it does! Think I found why it isnt working on humans, maybe that'll help!

code/modules/mob/living/simple_animal/hostile/abilities_common.dm
NopemanMcHalt: Thank you! I'll tweak it and some other stuff shortly.

NopemanMcHalt:
Should be it! This is just an experimental one, however, and I have other version of beastmaster ready with some faction shenanigans.
Still, let people play with it around and hope for feedback, but I probably just will replace it soon(tm).

PR #1493, opened 2023-01-26
Improved Weaponsmith + Fixes

BigBadPossum:
Weaponsmithing is... useless, annoyingly useless. A 2 point trait which can be outpaced by five minutes of searching junk piles inside Nash. It serves next to no true mechanical purpose besides flavors. My hope with these changes is to give it a reason to exist and maybe encourage more crafter types then everyone rushing speed + HP + damage.

## Pre-Merge Checklist

## Changelog
:cl:
-Added a trait_weaponsmith define
-Added said define to the weaponsmith trait.
-Fixed an issue with hunting rifles being given twice by the weapon smith trait, corrected to hunting shotgun. 
-Commented out advanced explosive crafting since everything iit did was done by regular explosives crafting, giving it no purpose.
 -Fixed a missing descriptor to technophreak so it details that it gives T3 machine part recipes.
 -Added a wide selection of basic gun mods to the weaponsmith perk given its lack of use otherwise. 
-Fixed the caravan shotgun recipe producing hunting rifles instead of caravan shotguns when crafted.

--Removed due to code limitations
 _-Added in a check for faulty gun mods found on scavenged guns that allows them to be removed if you have the weaponsmith perk._
/:cl:

### DIFF
--- a/code/datums/traits/good.dm
+++ b/code/datums/traits/good.dm
@@ -40,7 +40,7 @@ GLOBAL_LIST_INIT(weaponcrafting_gun_recipes, list(
 	/datum/crafting_recipe/ninemil,
 	/datum/crafting_recipe/huntingrifle,
 	/datum/crafting_recipe/n99,
-	/datum/crafting_recipe/huntingrifle,
+	/datum/crafting_recipe/huntingshotgun,
 	/datum/crafting_recipe/m1911,
 	/datum/crafting_recipe/varmintrifle,
 	/datum/crafting_recipe/autoaxe,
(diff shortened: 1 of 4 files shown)


code/datums/traits/good.dm
```diff
@@ -622,6 +633,8 @@ GLOBAL_LIST_INIT(former_tribal_recipes, list(
 		return //welp
 	QDEL_NULL(our_tongue.lick_bandage)
 
+// This does the same thing as basic explosive crafting by giving basic_recipe and adv_recipe. -Possum
+/*
 /datum/quirk/advanced_explosive_crafting
```
Superlagg, line 638: As this removes a trait, can you test to be sure this won't cause any issues with people who had the quirk before the PR? I forget what'd happen, I think I made it just reset your quirks if an invalid entry happened to be in there. Not a problem if it does this, I can just make an announcement and tell people to expect having to redo their quirks.

BigBadPossum: I tested it on a local server and discovered that only people who have the perk will need to manually hit "reset traits" to remove the trait after its been commented out.

BigBadPossum:
The function for allowing faulty mods to be removed has been deleted from the PR as after excessive and extensive testing and even getting other coders to look at the issue we could not find a way to get the function to work. Our only hypothesis is that their is a code limitation that makes it either impossible or requires work that is beyond the scope of what I wanted with this PR.

Superlagg:
> The function for allowing faulty mods to be removed has been deleted from the PR as after excessive and extensive testing and even getting other coders to look at the issue we could not find a way to get the function to work. Our only hypothesis is that their is a code limitation that makes it either impossible or requires work that is beyond the scope of what I wanted with this PR.

That's alright, I'll take a look later and see what's up!

Superlagg (approved):

PR #1642, opened 2023-03-06
Map coordinate system? Pog??

GremlingSS:
## About The Pull Request

Just did this as a sudden urge of energy after my panic attack from today.
Be careful, this one's fresh right out the oven.

Force X and Force Y and Force Z have been added to the maps traits, these are a **MUST** when it comes to adding in crosslinked maps.

but yeah it does exactly as the tin says, so now you can actually map whatever the fuck you'd like on any z level, any area of map, and wherever the fuck you'd like, have fun!

Though I will warn, there can't be any gaps in the map, they all have to be linked, hence the name crosslinked. If you don't want it to be part of the transition system, don't add "Cross" to the map in json. :)

## Pre-Merge Checklist
- [x] You tested this on a local server.
- [x] This code did not runtime during testing.
- [x] You documented all of your changes.

PR #1651, opened 2023-03-06
Ai cheeky optimisations

GremlingSS:
## About The Pull Request

In reality this looks stupid and small, but...
it practically cuts down the CPU resources for mobs from 0.000 to 0.000 (from testing), but, I can confidently say that the AI tries to list all targets in half the amount it did before. It's p good. Would recommend.

Only a SLIGHT optimisation, but I hope it makes mobs run a lot smoother in general because of this.

Removes a very important feature! Radiation from goop, it called every single gametick and took a lot of processing power, was very stupid. We'll look into implementing radiation a bit differently in the future if we decide to go down that path more <3

## Pre-Merge Checklist


GremlingSS:
You need to merge #1642 first too, woops.

Tk420634 (approved):

PR #1699, opened 2023-03-13
Removes atmos

GremlingSS:
## About The Pull Request

Wild, huh?

Yeah, removes atmos.

### Big Current task list:
- [x] Make sure nothing runtimes during normal gameplay
- [ ] Rework how fire works.
- [x] Rework the breathing procs in life to make you breath normally and then suffocate if in crit.
- [x] Cry and pee and maybe even- uhh, cry more maybe

# You may be asking in a fatmagic voice: "Why?"
###### ref: https://youtu.be/EmrYl_I_dAY?t=837
We hate atmos, all my homies hate atmos.
No, for real this can open up a lot of memory for the game to use in the future. The only thing atmos might be useful for is vaults but...eh... It's costs outweigh the benefits.

We'll detach from auxmos, citadel's way of handling atmospherics using rust libraries.

A small amount of performance overhead is provided allowing more time for mobs to be prioritised as they are our main bottleneck in performance from the looks of things currently.

## What does this do for the players?

Fire will have to be reworked. Meaning the way how fire currently works which is using obj/effect/hotspot on a turf and is then related with "active_hotspot" for handling atmos, we'll instead detach it entirely from the turf.

Currently in my design brain I have it as making it a smoke effect, resprited with it's own code.
I'm going to be reworking smoke effects too to be handled by a subsystem as we plan on using this for other things like stinky gasses and, well, fire!

Unfortunately, because we're removing the option for you to use your lungs in a meaningful way, lungs are now fluff, meaning they're just there for lore, and provide no oxygen for your brain, making you stupid, yeah, big dummy, just like me :)

This also causes the issue where anesthetic or gas tanks won't actually work, but we plan on making it so gas masks prevent you from being effected by stinky smokey gasses from being emitted into the atmosphere.

## Pre-Merge Checklist


Tk420634:
Alright, things we need to work on.

-Heat Snap Status Effects
-Cold Snap Status Effects
-Cold Blooded Trait Rework or removal

PrettyPsychoForAWolf (approved):

GremlingSS:
It might be worth just straight up merging this by next week if it works out well, the fire is in a stable spot to not cause danger, it's kinda funny too if you just catch fire for a second.

Tk420634 (approved):
https://www.youtube.com/watch?v=18CGW9P5Y9M

PR #1842, opened 2023-04-01
Adds a horrible april fools """""prank"""""

Superlagg:
## About The Pull Request
Lets admins turn every gun and magazine into american 180s. Also makes every gun and magazine runtime when they're spawned and fuck up the garbage thing.


GremlingSS (approved):
Lmao

PR #1867, opened 2023-04-04
Removes junk and adjusts weight for unique loot drops

Coolgat3:
## About The Pull Request
There's about 20 or so of those spawners out in the world per round. some absolute junk put in there only as a joke was moved to the more common spawners instead, Weights for the most common of the unique weapons were reduced, from 20 to 15 so the gap of rarity between the most common of these weapons to the second most common is less. This should hopefully make the weapons with 1 weight appear more than once per 2-5 IRL days.

We all clear high risk loot spots for high tier items, seeing a deathclaw spear pop in or having looted a spot with 3 of these spawners 10 times but never having found anything with a weight of 5 really pains and I'd love to see the more rare weapons appear at least some times.

## Pre-Merge Checklist

## Changelog
:cl:
tweak: changed weights of unique loot so the rarer pieces have a bit more weight proportionally, while the code is picking which weapon will the spawner spawn.
/:cl:


Tk420634:
Closed, good idea poor implementation.  We'll work out a system that's more solid internally.

PR #1983, opened 2023-04-17
Cannibal AntiMeta PR

AllCorruptingWyrm:
## About The Pull Request
AKA; I can not believe people powergamed this item to valid hunt. Adds an OOC note to the perk on spawn and removes the medical text.

## Pre-Merge Checklist

## Changelog
:cl:
tweak: tweaked the perk.
/:cl:

### DIFF
--- a/code/datums/traits/good.dm
+++ b/code/datums/traits/good.dm
@@ -100,12 +100,11 @@ GLOBAL_LIST_INIT(energyweapon_crafting, list(
 
 /datum/quirk/horrifying_tastes
 	name = "Cannibal"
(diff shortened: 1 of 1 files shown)


FuzzyFuzlet:
the change to the description of the quirk looks a tad redundant and wordy. medical_record_text is null by default and most traits you cannot see on another person without admin powers if they don’t have medical record text. so while I grant that removing that var is probably for the better, the second change is unneeded

HanSolo1519:
There's a system in place to hide antag quirks (or I might be mixing that up with antag moodlets, I know the traitor moodlet doesn't show on scanners)

I'll get back to you on that, there's more graceful ways to go about it then an OOC rule-note, but this is a good first step. Don't block merging the PR for this, if I remember too late I'll make a new one.

Tk420634 (approved):

PR #2177, opened 2023-05-05
Removes ghoul glow and changes their stamina recovery

Essiuun:
## About The Pull Request
Removes glow from ghouls and gives them more stamina regen so they TRULY never get tired.

## Pre-Merge Checklist

## Changelog
:cl:
commented out the ghoul glow, and upped their stamina regen. Yes, it's 1000 now, but they can never get tired. (Sprinting used to make them tired so that's why I changed it, yeah)
/:cl:


Tk420634 (commented):
Denied.  Ghouls will need a concerted overhaul by the normal team, please don't try to tweak them like this right now, especially not without running it past the entire team, preferably in voice.

PR #2419, opened 2023-05-26
In code Implementation for mob pathfinding.

gozulio:
## About The Pull Request
This commit makes mobs use an in-code implementation of AStar in order to path find.
It should fix issues with mobs getting stuck trying to path into diagonal squares they can't reach. However this comes at the cost of efficiency. An in-code implementation of AStar is always going to be less efficient then the one built into byond. I'll end up working on a way around this but for now this needs to be tested to see how bad that efficiency loss actually is. The range is limited to ten squares, which is more then the players view range.

Other minor additions:
 - Added a chance for mobs try and melee attack when they try to retreat. Radroaches specifically will attack more sporadically rather then getting stuck never attacking as they move in and out of range.

## Pre-Merge Checklist

## Changelog
:cl:
code: Hostile mobs now use an in-code implementation of AStar to path find.
fix: Radroaches sometimes never attacking as they move in and out of range.
/:cl:


gozulio:
This PR should be ready, Unless someone finds another bug.

Tk420634 (approved):

PR #2477, opened 2023-06-07
no more hilariously op shit for basically free (removes slimecores from shop)

jinxynii:
You don't need hilariously overpowered xenobiology stuff at round start for nearly nothing.


Tk420634 (approved):
Ur a slimecore (type beat)

PR #2490, opened 2023-06-08
Fixes mob post-nap grogginess

Superlagg:
## About The Pull Request
Mobs will now actually fully wake up and act like mobs after they wake up.

Also makes them more aggressively go to sleep.


Tk420634 (approved):
go to bed eepy, wake up HANGRY

PR #2508, opened 2023-06-12
Reworks recoil into something I can understand

Superlagg:
## About The Pull Request
Recoil is now a much more robust and unlaggy subsystem.

Recoil cursor should be a million percent less laggy.

Redid the TGUI gun window thing.


Tk420634 (approved):

Superlagg:
imma just merge this k

PR #2548, opened 2023-06-25
Sleeping Carp fix and "nerf"

jinxynii:
Gives the no drugs trait to sleeping carp, as it was intended from the getgo. Halves adrenaline implant's reagents while also giving it only 1 use. You don't need 3. The only reason this ever even gets used is for sleeping carp, to instantly regain all of your stamina back-- 3 times.

For the same reasons I removed the slime cores, this is being nerfed. If it continues to be a problem, I will hit it again.


Superlagg (approved):

CloverIsLucky:
Imagine using sleeping pussy, based PR nerf.

PR #2593, opened 2023-07-03
Slight refurbished power armor tweak

RinPin:
## About The Pull Request

Makes the refurbished power armor have a speed that is slightly slower than regular PA, at 1.2 instead of 1, since at the moment, it's just as slow and bad as  salvaged despite all the trouble it takes to get the servos restored + the training to wear it, while still keeping the salvaged armor values.
## Pre-Merge Checklist
- [Y ] You tested this on a local server.
- [Y ] You documented all of your changes.

## Changelog
RinPin :cl:
tweak: refurbished PA now moves at slightly slower than PA speed

/:cl:


RinPin:
i made another PR that is the exact same as this one, but adds the new speed to all the refurbished armor, no point to this one

PR #2614, opened 2023-07-09
Second Wind

Superlagg:
## About The Pull Request
Adds in a way for players to stop being dead if they want to.

Everyone starts with a Second Wind. If they die, they can revive their body at any time, granted it is in one piece and not handcuffed.

When you revive, you'll need to stay alive for 2 hours to get that second wind back.

If you die again without a second wind, you can revive yourself, but it'll be the last time you can revive yourself for the rest of the round.

isnt ready yet


LiteralMushroom:
yeah uuuuUHHHHH make it so you just. Can't revive yourself without second wind because the "lose it forever if you use it twice too close together" mechanic is weird and unnecessary (please)

unless the description is outdated

ThatGuyFromThatPlace:
> yeah uuuuUHHHHH make it so you just. Can't revive yourself without second wind because the "lose it forever if you use it twice too close together" mechanic is weird and unnecessary (please)
> 
> unless the description is outdated

100% agreed with this
3rd wind should be, IMO, either not exist
Or be a positive trait with a cost
And not be called "Third wind" because that's dumb
Last gasp sounds so much better

Tk420634 (approved):

PR #2647, opened 2023-07-16
ITS BULLET HELL IN HERE HOLY SHIIIIT

jinxynii:
bit too many shots lmfao


Tk420634 (approved):
Denied, Touhou is the goal

PR #2650, opened 2023-07-17
Linter fix

GremlingSS:
Augh, attempts to fix linters cause I hate seeing red X's

For future reference:
Do not put mobs near other mobs not in the same faction, this leads to lots of errors.

- [x] When this is ticked, it do be ready


Tk420634:
bruh

GremlingSS:
drop-the-mic-obama-mic-drop

Tk420634 (approved):
Big stinki

PR #2662, opened 2023-07-20
Fixes Excavator PA

RinPin:
Gives Excavator PA the same laser armor the helmet offers, which is meant to be bad rather than non existent.
Also gives it a flashlight like the other PA helmets (it's a mining tool after all)

## Pre-Merge Checklist
- [Y ] This code did not runtime during testing.
- [Y ] You documented all of your changes.

## Changelog
Rinpin :cl:
fix: Excavator PA has laser armor
/:cl:


Tk420634 (approved):
denied for smelling like cut grass

PR #2716, opened 2023-07-29
Tweaks Ghoul Melee Alot

DiscordWizard:
## About The Pull Request

This makes it so ghouls just deal 20% less damage instead of dealing flat out 20% of normal damage.

Right now ghouls can grab the best melee perk in the game, the most damaging weapon in the code, dual wield it and hit someone with the grant total force of a random guy with a rusted toolbox. Considering ghouls already don't have the free out of jail card that critical condition is, are slower by default and heal less from stimpacks, with rad immunity and healing being the one thing going on for them, I feel like making it impossible for anyone to make a melee ghoul gimmick is pretty unnecessary.

If you want an example, a warhoned axe deals 65 damage when wielded. This trait makes it so the ghoul deals 13 instead. If you take big leagues to attempt to compensate, you deal the same damage as a literal toolbox.

Just bonk the guys with a sledgehammer or a rapid fire weapon if you're worried about PvP balance, believe me they fold like a pancake at that point.
## Pre-Merge Checklist
- [ x] You tested this on a local server.
- [ x] This code did not runtime during testing.
- [ x] You documented all of your changes.

## Changelog
:cl:
balance: rebalanced ghoul damage
/:cl:


RinPin:
oh wow, 20% damage output was whack, this is a good change for sure.

code/_onclick/item_attack.dm
```diff
@@ -111,7 +111,7 @@
 	var/force_modifier = 0
 	if(force >= 5)
 		if(HAS_TRAIT(user, TRAIT_PANICKED_ATTACKER) || HAS_TRAIT(user, TRAIT_GHOULMELEE))
-			force_modifier = (-force * 0.8) // You do 20% damage cus ur scared
+			force_modifier = (-force * 0.2) // You do 80% damage cause you're a walking corpse
```
Superlagg, line 114: This would also make adrenaline do 20% less damage instead of 80% less damage. You'll want to separate this into two different trait checks.

Superlagg (changes requested):

code/_onclick/item_attack.dm
DiscordWizard: Oh my god I didn't even realise they paired it with some other trait, why even do that. I'll fix it tomorrow I need to sleep, thanks for pointing it out.

EtheoBoxxman:
I may be biased, since I want a ghoul crusader archetype, but I feel this is more than reasonable. 20% less damage instead of 20% total damage for melee when its already high risk to go melee as a ghoul feels plenty reasonable.

I need this in my life. u wu

Superlagg (approved):

PR #2835, opened 2023-08-22
Preacher Love

BigBadPossum:
## About The Pull Request
For awhile now I have observed that the preacher role largely gets the short end of the stick compared to a standard wastelander. For one their non-weapon option is a duplicate null rod (they already spawn with one) that is such a pathetic weapon that it barely deserves to be called one. Another issue I noticed was that preachers did not get a Nash radio despite, in all but code, being a nash job. As such, I decided to put my coding where my mouth is and made this PR after a brief ingame discussion.

Major Changes
The preacher can now select some decently balanced weapons in their load out, all of which being subtypes of the null rod. This includes some moderately universally decent weapons (claymore and all subtypes), to defensive options (staff/rosary), to thematic (katana/cult blade) to slightly meme worthy with the dimensional blade. These mildly unique weapons are by no means very powerful, the strongest is still on par with the starting simple sledge, and do not invalidate any existing job (the rosary takes quite awhile to heal and can be considered supplemental at best).

Goal
By making preacher a proper citizen, giving them some unique options, and shortening the job to only 1 person this should, I hope, create a more thematic option for the few preacher players around. As there are now thematic items for occult priests, holy priests, war priests, healing priests, and even 'techy' priests.

One Issue
I am not sure if its because I am using non-tg hosting daemons but for some reason I could not boot a test server at all. It simply would not load properly. Thus, this will have to be tested. I can at least assure it compiles and should in theory work. **For testers: Spawn as the preacher, check if you have the nash radio, active your spiritual device kit and see if it contains more than the null rod.** If that is right, we're good to merge.

## Pre-Merge Checklist
- [] You tested this on a local server.

## Changelog
:cl:
add: The spiritual device kit now contains multiple options instead of the currently useless null rod. Items include a war hammer, claymore, shield staff, bo staff, rosary, cult blade, and multi-blade.
del: Removed the default null rod from the preacher outfit/bag.
tweak: The preacher role has also been dropped to 1, given it is a singular chapel and thus a unique position.
balance: Priest hammer now has roughly identical stats to the starting sledgehammer as to not be outright outclassed.
fix: Preacher should now spawn with a nash radio.
/:cl:


thingpony (approved):

Brimcon:
Looking this over, I believe that the nullrod itself transforms into some of these weapons. They do however require their own tweaks because they have default '/tg/' values. One thing we will look into is a complete preacher revamp/pass. We will look further into this PR all the same. Going to keep it open for discussion and further tweaks.

Tk420634:
> Looking this over, I believe that the nullrod itself transforms into some of these weapons. They do however require their own tweaks because they have default '/tg/' values. One thing we will look into is a complete preacher revamp/pass. We will look further into this PR all the same. Going to keep it open for discussion and further tweaks.

Alternative argument, you're a smelly cat.  Also yeah, what brim said.

BigBadPossum:
Hah, now here is something amusing. When I originally tried preacher I tested the null rod bia the old right click mode and it did not adjust. Turns out it *can* be adjusted, through the use of activation. Amusing over sight, I will close this PR then and make a new one to adjust the radio.

PR #2841, opened 2023-08-22
Preacher Love 2: Electric Boogaloo

BigBadPossum:
A rebalance and adjustment to the preacher job that, as it stands, is largely weaker and somewhat useless compared to wastelander. First and foremost, making them an actual nash citizen and rebalancing a few of their items to function properly when compared to the balance of Fallout versus the old /tg/ values.

This is a better researched and balanced PR of my prior closed one, after taking the time to test out how null rods work.

Major Changes
Preacher weapons derived from the null rod now have better stats, with each 'on par' item matched directly in the code to ensure they are fit to be useful compared to other starting weapons.

Goal
Giving the preacher something, even if it is barely a bonus given this only brings them on somewhat equal footing compared to most, rather than simply being inferior. It also fixes a long time issue (I think) of preachers not being Nash citizens.

One Issue
I am not sure if its because I am using non-tg hosting daemons but for some reason I could not boot a test server at all. It simply would not load properly. Thus, this will have to be tested. I can at least assure it compiles and should in theory work. **For testers: Spawn as the preacher, check if you have the nash radio.**

Pre-Merge Checklist
[ ] You tested this on a local server.
[ ] This code did not runtime during testing.
[X] You documented all of your changes.

Changelog

🆑
balance: 
-The null rods default damage is now 30, equal to a war club, this effects the base line damage of most null rod subtypes.
-The holy claymore now has stats identical to the spatha/forged claymore.
-Ripper is renamed to the chain sword, given a new description, and damage remains unchanged (was identical to the ripper already)
-Multiverse sword damage cap upped to 40 (randomly rolls 1-40 per hit, meme weapon already and easily the worst)
-Scythe and by extension the vibro blade now have equal stats to the 2 handed axe, including swing slow down.
-The relic hammer now has a base damage of 30 and wielded damage of 65, equal to the sledgehammer.
-Changed clown to mutant in the bo staff description.
-armwraps of unyielding resolve damage bonus upped from 2 to 8. It is still directly and immensely inferior to the swamp star wraps as it lacks rapid and has a smaller bonus, but allows range weapons and can be removed.

fix: Preacher should now spawn with a nash radio.
/🆑


Tk420634 (approved):
Gonna see what the hold up on this is from the other devs today, rest assured I'll be bapping them with sticks.  idk why I haven't been told this is good or not yet.

Brimcon:
The weapons the Null Rod turns into is missing the attack speed variables.

BigBadPossum:
It should be the default of whatever is taken from the above parent if the null rod does not have it. I can add it if needed.

Brimcon:
Please do, All melee weapons will need their appropriate/equivalent speed variable for the future.

BigBadPossum:
> Please do, All melee weapons will need their appropriate/equivalent speed variable for the future.

New changes. Also, I may have found a possible exploit with old weapon code, I am going to test it and report my findings if it is, I don't want to say what it is lest it get abused in the mean time.

-Most weapons were given a speed total modifier.
-After reviewing speeds to specific weapons some were further rebalanced.
-Add the "can parry" tag to the base null rod.
-On further examination the unyielding resolve gloves were vastly buffed, bringing them in line with the mace gloves.

Brimcon:
Everything looks good.

PR #2892, opened 2023-08-29
Mobs no longer fight each other

Superlagg:
## About The Pull Request
All the mobs got together and decided to put aside their differences and focus on the real problem: you. Specifically you. 

Mobs no longer fight other mobs.


Tk420634 (approved):
In the words of gob.   "Fuck the players"

But for real, this is good QOL.

PR #2923, opened 2023-08-31
You can now wear collars and necklaces on the mask slot!

OverDriveZ:
## About The Pull Request
It's now possible to wear some simple necklaces and collars on the mask slot. I think this could be interesting not only from an aesthetical point of view, but it could also incentivize players to get the 'masked mook perk'. Like for example, a mercenary that is deeply attached to their dog-tags.

On a side note, it's not possible to activate mood debuffs/buffs when wearing collars as a mask. Meaning that is not possible to stack two positive buffs using two collars. One of them simply becomes cosmetical. 
immagine
immagine

## Pre-Merge Checklist

## Changelog
:cl:
add: This new cool thing that allows YOU to wear collars on the mask slot.
/:cl:

### DIFF
--- a/code/modules/clothing/neck/_neck.dm
+++ b/code/modules/clothing/neck/_neck.dm
@@ -10,7 +10,7 @@
 
 /obj/item/clothing/neck/equipped(mob/user, slot)
 	. = ..()
-	if (slot == (SLOT_NECK) && istype(user))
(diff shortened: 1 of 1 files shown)


Tk420634 (approved):
D R I P

PR #2938, opened 2023-09-02
Fixes Spirit Boards

EtheoBoxxman:
Reverts the changes to spirit boards
Creates a new parallel to spirit boards, referred to as a 'spirit hole', so fenny can have his own unique tribal version to play with.

Also adjusts druids' seed spawns and fixes their book.


Tk420634 (approved):
https://tenor.com/view/brainlet-brainlet-bike-ding-ding-ding-gif-20767472 me rn

PR #3008, opened 2023-09-08
Scrubs suicide from the code

Superlagg:
## About The Pull Request
If you want to commit die, you'll have to do it the old fashioned way


Tk420634 (approved):

LiteralMushroom:
ok but why did you remove the WORD suicide and not just the actual mechanic?

PR #3087, opened 2023-09-18
New Job, Quirk Tweaks, and new Quirks

EtheoBoxxman:
Replaces telecomms agent ghost role with a proper limited job role

Also gives their lil post a matrix in case they wanna leave

Should be fine after a triple check

Also re-fixes some perk value fixes that got undone somehow
phobia cat from -1 to -11 and trekker from 77 to 55

Adds tiered surgery quirks

Re-Impliments dark phobia and light phobia

Renamed Light phobia to Light Sensitivity, which now is more severe than before in that if you are in a bright area, you will now suffer from severely blurred vision.

The blurred vision particularly can be circumvented with sunglasses, but other negative effects remain.

Adds a new 'Wild Shape' quirk, that gives you a spell to let you select a simpler form and transform back and forth at will. Want to be an eevee in your free time? Now you can be.


Tk420634 (approved):
Let's fucking GOOOO

PR #3096, opened 2023-09-18
two small fixes!!!!!!!!!!!!!!!!!

cebutris:
Fixes two things, although one is more of a tweak. The mailbird now properly flies away instead of laying dead on the ground, and also Primitive no longer prevents you from playing intruments


Tk420634 (approved):
ur butts primitive

https://www.youtube.com/watch?v=5ZG4iejbM5Y

PR #3097, opened 2023-09-18
dark vision and tribal carpet bombing

EtheoBoxxman:
makes dynamite tiny sized items
does 30 dmg, so the only way this can kill is if you piss off the whole tribe.

Will need a timer on sending birds, maybe five minutes, if this is used too much.

Also changes night vision to two perks
Dark vision - minor
dark vision - greater

2 and 4 tiles dark vision range, 22 and 44 point cost respectively


Tk420634 (approved):
Pearl Harbor moment, ngl

PR #3099, opened 2023-09-18
fixes the locksmith kit

cebutris:
smol fix PRs amirite. eventually I'll do something bigger

Now the kit has the correct kind of lock and keys. I don't actually know what the ones it had before are used for.


Tk420634 (approved):
I'll lock you

PR #3132, opened 2023-09-21
axes the proton axe

cebutris:
So I was investigating the source of the proton axe's log spam, hoping I could come to some sort of solution to make it chill a little, but I found out that the effect was actually massively broken. The EMP effect did not check for proximity or attack speed, meaning you could spam a targeted, one-tile EMP attack on anything you could see as fast as you could click. I don't really want to go through the effort of fixing this, but it's obviously not okay in it's current state

So now the shitty fireaxe clone is a little faster


Tk420634:
Hmm, we'll do this for now but lets see if we can get someone with actual brains (not us) to look at it and make it work right.  So, temp fix.

Tk420634 (approved):

PR #3138, opened 2023-09-22
Sound Indicator Animal Crossing style [don't merge, I need help with one thing]

OverDriveZ:
## About The Pull Request
It adds a new sound indicator option, it allows the player to talk like an animal crossing character!
Like this:

https://github.com/ARF-SS13/coyote-bayou/assets/64517916/2ce9b043-413e-49c6-a062-0b4a23bee38f

We could have definitely ported a better way to do it from somewhere else, but I really didn't want to get rid of the previous coder's effort, so I made it work with what we already had.

## Pre-Merge Checklist
- [x] You tested this on a local server. god, yes I did so many times

## Changelog
:cl:
add: Adds animal crossing style speech.
/:cl:


OverDriveZ:
I have to postpone this, it's... making some errors I'm not sure how to fix

code/modules/keybindings/keybind/communication.dm
```diff
@@ -30,6 +30,7 @@
 /datum/keybinding/client/communication/say_with_indicator/down(client/user)
 	var/mob/M = user.mob
 	M.say_typing_indicator()
+	M.play_sound_typing_indicator()
```
Superlagg, line 33: Just have `say_typing_indicator()` run this proc, would make it so you dont need a global var!

Superlagg (approved):

code/modules/mob/say.dm
```diff
@@ -1,17 +1,25 @@
 //Speech verbs.
 // the _keybind verbs uses "as text" versus "as text|null" to force a popup when pressed by a keybind.
+var/global/message_length = 0
```
Superlagg, line 3: Global vars here are made through `GLOBAL_VAR(varname, value)`, though like I said above, isnt really needed to be a global.

Superlagg (changes requested):
Found the reason for the runlinters, and how to fix it!

code/modules/mob/say.dm
OverDriveZ: thank you, changing it now! Though it will have to wait a little more as I'm trying to add a global preference function to mute all vocals

code/modules/keybindings/keybind/communication.dm
OverDriveZ: there's an issue with this, using a single proc makes it impossible to begin saying a new sentence until all sounds have been played

GremlingSS (approved):
looooks good

PR #3150, opened 2023-09-23
Lets players eat mobs, sort of

Superlagg:
## About The Pull Request
You can eat just about any simple mob! Its gotta be dead first, or have a player currently in it. 

Also, simplemobs can eat *you*, but only if theres a player in them. Do note, they can't hold the emote, so you'll have to do the backup thing, where you grab someone and type *vore.


Tk420634:
I'll chatter with jess about this one later, since they lack hands it will probably be fine.

Tk420634 (approved):

PR #3174, opened 2023-09-26
Disabled the theme warp crystal

BlackMajor:
Title.

Disables the theme warp subtype of the anomalous crystal, after turning the entire map into lavaland with it (whoops)


Tk420634 (commented):
We're just gonna pull it from the drop table, we'll test it later to make sure it doesn't do other things we don't want.  Big surprise, things that worked for tg years ago don't work for a server that is a totally different environment.

BlackMajor:
> We're just gonna pull it from the drop table, we'll test it later to make sure it doesn't do other things we don't want. Big surprise, things that worked for tg years ago don't work for a server that is a totally different environment.

You want me to tweak the PR to exclude it from the subtypes then?

Tk420634:
Naw, we got it.

PR #3196, opened 2023-09-28
Makes raging boar ANGRIER

RinPin:
as it stands, raging boar is very weak compared to the other martial arts, this gives it a buff to make it more viable, by adding a flat damage boost to the punches, while making it scale worse than the beserker's multipliers, reduces slightly the stamina cost of the tackles, and gives it a weaker stamina/ stun modifier than the other martial arts (which was previously lacking)

## Pre-Merge Checklist


Tk420634 (approved):
As always, with any MA balance change, be ready to tweak it.  I'm sure there is a place that would be perfect for them but we've all seen how easy it is to over/undertune the things.  I really do love the raging boar vibe though, ngl.

PR #3216, opened 2023-09-29
Refactors pornhud - READY

Superlagg:
## About The Pull Request
Nothing like ripping out some of my worst procs ever and replacing them with some sickass subsystemry. The whitelist is actually responsive, it doesnt make literally everyone update their everything always all the time (maybe), and it has some nice profiling attached to it (sort of).


Tk420634 (approved):
My god, let it be so.

PR #3220, opened 2023-09-29
Makes it so unit tests only load one map (should reduce checks by a c

GremlingSS:
…onsiderable amount)

## About The Pull Request

Doesn't affect actual gameplay, just makes it so github checks are like, considerably faster.


OverDriveZ:
this is based as fuck

Tk420634 (approved):
*changes ur one map to be 2000x2000

PR #3242, opened 2023-10-02
adds in naomi reids

cebutris:
Adds in the basic sweaters to the loadout, so you can run around without pants in style.

Also adds some dumb belt jokes????????


Tk420634 (approved):
I'm sure she'll need nerfing.

PR #3245, opened 2023-10-02
Distinct trait

cebutris:
Adds in a new neutral quirk!

Distinct will prevent masks and helmets from concealing your identity. Useful if you want to wear a mask but still want people to be able to read your flavor text


Tk420634 (approved):
Horny

PR #3280, opened 2023-10-04
Revert crate crafting changes

Varoxus:
## About The Pull Request
Reverts the increased costs to craft crates and instead makes them have next to no export value. There was no reason to do this in the first place when you can just make the export value lower to combat any exploiting. The only group you're hurting by increasing the material cost to create crates is general wastelanders.
## Pre-Merge Checklist

## Changelog
:cl:
tweak: reverted crates to their old crafting values, made them worth a lot less.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/sheets/sheet_types.dm
+++ b/code/game/objects/items/stacks/sheets/sheet_types.dm
@@ -83,7 +83,7 @@ GLOBAL_LIST_INIT(metal_recipes, list ( \
 	null, \
 	new/datum/stack_recipe("shelf parts", /obj/item/rack_parts), \
 	new/datum/stack_recipe("closet", /obj/structure/closet, 2, time = 15, one_per_turf = TRUE, on_floor = TRUE), \
-	new/datum/stack_recipe("trash bin", /obj/structure/closet/crate/bin, 30, time = 15, one_per_turf = TRUE, on_floor = TRUE),\
+	new/datum/stack_recipe("trash bin", /obj/structure/closet/crate/bin, 10, time = 15, one_per_turf = TRUE, on_floor = TRUE),\
 	null, \
 	new/datum/stack_recipe("computer frame", /obj/structure/frame/computer, 5, time = 25, one_per_turf = TRUE, on_floor = TRUE), \
 	new/datum/stack_recipe("modular console", /obj/machinery/modular_computer/console/buildable/, 10, time = 25, one_per_turf = TRUE, on_floor = TRUE), \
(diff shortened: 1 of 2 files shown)


Tk420634:
1) Remove the change to the crate value

2) The original crate crafting costs are no good. However, you can change the values from what they currently are. (metal has to be AT LEAST 25/crate, plasteel at least 10 or more, wood minimum of 20)

Both of those, or please stop wasting my time. 😃

Varoxus:
> 1. Remove the change to the crate value
> 2. The original crate crafting costs are no good. However, you can change the values from what they currently are. (metal has to be AT LEAST 25/crate, plasteel at least 10 or more, wood minimum of 20)
> 
> Both of those, or please stop wasting my time. 😃

Only reason I wanted them changed back is because if you're using a crate to drag things around with you and it gets destroyed while you're out in the field... you now can't craft another crate to replace it since crates drop the same number of materials when they break. If it's such an issue then I won't bother, but you didn't have to be so hostile about it.

Kelprunner:
So, like... You answered your own issue here? Just make it so that crates, when destroyed, drop the new crafting cost in material. Or close to it, at any rate.

Besides, you can now carry up to ~60 sheets in a box or toolbelt, or 100 in a bag. Like, even without that fix it really isn't hard to get the materials to make a new crate unless you're **deep** in a dungeon.

Tk420634:
> So, like... You answered your own issue here? Just make it so that crates, when destroyed, drop the new crafting cost in material. Or close to it, at any rate.
> 
> Besides, you can now carry up to ~60 sheets in a box or toolbelt, or 100 in a bag. Like, even without that fix it really isn't hard to get the materials to make a new crate unless you're **deep** in a dungeon.

You know, a place a shopkeep shouldn't be. 🙊

Varoxus:
As a ruin-delving main and not much of a shopkeeper, I can't really speak to any exploits. I do think making crates suddenly 6x more expensive simply because shopkeepers are exploiting it (Might be other reasons, but that was the main one stated.) seems to solve the problem at the expense of everyone that's not a shopkeeper. 

"Like, even without that fix it really isn't hard to get the materials to make a new crate unless you're **deep** in a dungeon."

-Main issue with that is that most crates there were changed are made of plasteel, which you'd need to mine for and then deposit into an ORM to even be able to get new sheets of, since plasteel spawns aren't guaranteed. If it's an issue of balance then I'll leave it as is, I don't know exactly what people are doing so I can't say if it'd be a good idea to increase the material drops or not.

PR #3285, opened 2023-10-05
Fallout 13 Energy Overhaul (DO NOT MERGE, NOT UP FOR DISCUSSION!)

Kelprunner:
## About The Pull Request
Does just all the things with F13 energy code.

Small Energy Cells, Microfusion Cells, Electron Charge Packs, Ultracite cells, and Microfusion Breeders are made equivalent to the stock batteries for consistency's sake. Weapon power drain (for F13 weapons using these cells) has been adjusted accordingly, and some weapons were buffed/nerfed. Also touched some weapon upgrades, crafting, and loot stuff to account for all of this.

## Pre-Merge Checklist

## Changelog
:cl:

- SECs are equal to T1 cells
- ECPs are equal to T2 cells
- MFCs and MFC Breeders are equal to T3 cells, MFCB is not actually self-charging
- Ultracite cells are equal to T4 cells
- Alien powercell is equal to a hypothetical T5 cell; still can't be charged
- Nerfed and buffed various energy weapons, notably buffed the AER-14 slightly and nerfed the Autoshock/Twin-Shot carbine to more sane levels; too many tweaks to count, but the comments _should_ be accurate for shots-per-gun now; may need a second pass as I had made some ECP guns drain like they used MFCs. Not sure if I got all of them, but we'll see
- Added a recycled SEC, ECP, and MFC battery; identical to normal batteries but can't be recharged
- Hostile robots now drop recycled batteries so MFCs aren't comically easy to get
- Lasers can no longer be plugged into weapon rechargers (affects even non-F13 ones, I need to go in later and manually fix the classic lasers; this was done to stop from being able to charge non-chargeable cells)
- Self charging lasers should no longer be able to have their cells ejected or hot-swapped, so their self-charging is a tradeoff and not an objective upgrade
- Tweaked the Damage+ laser mods to not only be more consistent, but actually take into account what they can pair with; largely been nerfed with harsher RoF penalties on top of hurting 1h recoil more (most 1h guns can be quickdrawn, and giving them more power is iffy so...)
- Recycling cells via crafting makes them into the worn-out version of a cell and can't be recharged. This makes actual ammo cells and shoddy cells actually valuable if you like recharging them. Can still turn shoddy cells into magazines if you really want to, however 
- Adjusted the description of some items to make their function more accurate or obvious
- Changed self-charging to a percent system; now you can make a gun recharge itself exactly how you want to instead of hoping 100 energy every 2 seconds wasn't OP or useless
- Changed the printing cost of SECs, ECPs, and MFCs in the science protolathe to actually require some material instead of being dirt cheap and consisting only of metal/glass
- After discussion, made shoddy energy cells 75% capacity instead of 50%

:cl:


Tk420634 (approved):
maintpurge

PR #3301, opened 2023-10-06
Changes bisexual alpha from 0 to 255

Superlagg:
## About The Pull Request
THe stat panel now supports bisexuality. Thought it did before, but I must have overlooked it.


Tk420634 (approved):
Blessed width saving the bisexuals from non-existing

PR #3313, opened 2023-10-07
Rejiggered Nash gate and Buttons

Krer-Grimpaw:
Turns out the inner gatehouse buttons rested on the exact same square as the outer buttons, making them redundant. Also changed gatecodes to be numeric only so if some chucklefuck nukes the controllers it may be actually FIXABLE

## Pre-Merge Checklist

## Changelog
:cl:
tweak: tweaked a few things
fix: fixed a few things
/:cl:


Tk420634 (approved):
Smelly but I'll take it

PR #3314, opened 2023-10-07
Smart inventory slot unsheathing [ready to merge]

OverDriveZ:
## About The Pull Request
Stupidly complicated smart equip/unequip, what is this supposed to do:
If the user is empty handed and triggers this function, then unholster an item with priority and with exceptions in the following order:
1-Backpack slot;
2-Armor slot;
3-Belt slot;
4-Holster contents, return ONLY firearms, all other items will be ignored;
5-Boot contents.

It also ignores people who wear their flashlights on these slots, it's smart, smarter than me.

There's one more twist to this, for example, if the user had previously unsheathed a revolver from a shoulder holster,
we clearly want the revolver to be re-sheathed in the previous location, this is why memory of item and origin of said item have been added.

There's one edit of this, if using quick equip with shoulder holsters, the cocking delay of the gun is close to 0, but before you freak out and call this change unbalanced, let's think about consequences of this.
First of all, in order to quick equip a weapon on your hands from the shoulder holster, it needs to follow a priority order, meaning that armor slot and belt slot need to be free of weapons. This implies that your handgun in your shoulder holster is your main weapon, there aren't many guns that can fit in this space to begin with too.
Second of all, committing to a shoulder holster means that you're not going to wear other also useful items such as the casing bag or the vietnam-war style flashlight, this on its own it's a big investment.

This pr also reactivates the quick backpack and quick belt equip, because they can be useful for QoL stuff. we can always turn them off if these becomes an issue, but benefit of the doubt.

## Pre-Merge Checklist

## Changelog
:cl:
add: Adds smart unsheathing.
/:cl:

### DIFF
--- a/code/modules/mob/inventory.dm
+++ b/code/modules/mob/inventory.dm
@@ -1,7 +1,3 @@
-//This variable is needed in order to remember the origin of the storage of an item.
-// GLOBAL_VAR_INIT(quick_equip_memory_item, 0)
-// GLOBAL_VAR_INIT(quick_equip_memory_origin, 0)
-GLOBAL_VAR_INIT(quick_equip_cowboy_delay_negation, 0)
 //These procs handle putting s tuff in your hands
 //as they handle all relevant stuff like adding it to the player's screen and updating their overlays.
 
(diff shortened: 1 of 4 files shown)


Tk420634:
immediatly merging ❤️

OverDriveZ:
Thanks to Rin, we now have an extra useful check

code/modules/mob/inventory.dm
```diff
@@ -1,3 +1,7 @@
+//This variable is needed in order to remember the origin of the storage of an item.
+// GLOBAL_VAR_INIT(quick_equip_memory_item, 0)
+// GLOBAL_VAR_INIT(quick_equip_memory_origin, 0)
+GLOBAL_VAR_INIT(quick_equip_cowboy_delay_negation, 0)
```
Superlagg, line 4: This should be a var on guns themselves, especially since we have access to the specific gun being equipped. Why? Global variables are generally used for things that need to have a certain state shared across multiple other things, like cached data, shared cooldowns, and such.

Superlagg (changes requested):
Very good PR, very solid, and from what I've heard very *intuitive!* It's something that I've wanted in the game for *ages* but never actually got around to. Just one thing I gotta ask about, the global var, see below.

code/modules/mob/inventory.dm
OverDriveZ: oh yeah, I don't know why I didn't do it like that to begin with. Allow me to change that

OverDriveZ: See Iteration VII

Superlagg (approved):

Superlagg:
Beaut!

PR #3353, opened 2023-10-10
fruit ninja update

RinPin:
## About The Pull Request

Adds a cosmetic version of the space ninja outfit to the loadout, lightly armored.
Adds 2 new weapons, the energy katana, which is a new type of plasma cutter and the hand crossbow, which is a small sized crossbow that can be fired one handed, at the cost of a very poor damage modifier.
also with intent of comboing with the hand crossbow, makes jagged arrows slightly less pathetic (they're still awful despite the numbers)
## Pre-Merge Checklist
- [Y ] You tested this on a local server.


RinPin:
There is no conflict, it's just bitching that i added new icons

Superlagg (approved):

PR #3430, opened 2023-10-14
Adds a sawed off hunting shotgun + crafting recipe

McRex007:
## About The Pull Request
Reusing the existing but unused sawed off pump shotgun sprite, implements a sawed off shotgun as a separate weapon. Normal sized, fits in holsters, but only holds 2+1 shots. Also hacked up an empty sprite. Balance wise, the sawed off widowmaker can fire 2 shots much faster, and doesn't need to be manually cycled, meaning operating it is as simple as, click click x click x click click in terms of shooting and reloading. More effort and a slower firerate for one extra shot and a whole load of badassery.

You can fire it one handed, but it turns out you can fire a lot of things one handed. I do remember at one point you were forced to wield things two handed, but that seems to have changed. The widowmaker and rebored widowmaker can also be fired one handed.

## Pre-Merge Checklist


McRex007:
I forgot to mention the sawed off pump has a T1 damage malus just in case it might end up being effectively a straight upgrade in peoples eyes

Tk420634 (approved):

PR #3461, opened 2023-10-16
Some artifact rebalancing

Superlagg:
## About The Pull Request
It makes your rtifacts not hurt you anymore, but some of them make you dizzy, so, uh, good. Not done yet.


Tk420634:
This is probably fine, Dan. Lets just crank the spawn amount and see what happens!

PR #3463, opened 2023-10-16
Adds tribal brute cream and fix aloe cream sprite (Not ready to merge yet)

McRex007:
## About The Pull Request
 Adds a cream you can make from horsenettle, and fixes aloe cream to use its proper icon state. Regular horsenettle cream causes the same effect of styptic powder and silver sulfadiazine where it hurts and makes you scream. The improved version just makes you a little uncomfortable, but only tribals know how to make it. 

It should be the case that all do, but because on my test server I kept spawning on the background with nothing but the belt loadout selector (except for radio operator for some reason and at that point I stopped bothering) all I was able to confirm was that the generic tribal traditionls booklet gives you the improved crafting recipe. If anyone can confirm that the title screen thing happens for them, why it happens or how to fix it, please let me know.

## Pre-Merge Checklist


McRex007:
Didn't fix the whole spawning on the menu bit, unfortunately, so I still can't test if tribals get the recipe for the improved one

Tk420634:
Dis good?

McRex007:
Yes, I think I fixed the spawning bug by fixing the comma. Might want to testmerge it just in case though.

Tk420634 (approved):

PR #3501, opened 2023-10-19
Fixes Rooftop Nomad Quirk Package

EtheoBoxxman:
Didnt give the tackling ability because Im a dummy and forgor to add the component codeblock


Tk420634 (approved):
Quackages are proving to be a bit more finicky than we expected.  Maybe instead of making them we should do a quirk var that checks 'if 1 & 2 quirk total = -1' or something idk

PR #3502, opened 2023-10-19
Fixes Snr Doc Medbeam

EtheoBoxxman:
It was fixed before by adding a loadout box, which was the most complicated way to fix it lmao.

Added it to the suit storage instead, to make it seamless.


Tk420634:
sneatho pls fix @EtheoBoxxman

Tk420634 (approved):

PR #3539, opened 2023-10-21
Nerfs the laser AK firing speed by a considerable amount and the damage by T1

RinPin:
## About The Pull Request

downgrades it's RPM from 300 (which was actually 600) to 100 (200) which hopefully will make it a LOT less insane than it currently is, alongside a minor damage decrease.

## Pre-Merge Checklist
- [Y ] This code did not runtime during testing.


Tk420634 (approved):
Lets do both this time, but in the future for stuff like this it's probably better to move one step at a time, don't want to hop over the potential Goldilocks zone

PR #3541, opened 2023-10-21
Buffs pellets and nonlethals against simplemobs

McRex007:
## About The Pull Request
 The purpose of this PR is to reintroduce the "better against mobs, worse against players" kind of ammo that 5.56/7.62 vs .223/.308 once provided, while also helping give nonlethals a place in PvP as a sort of "I got caught with my pants down and this is all I have."

With the exception of rubbershot, rubber bullets aren't touched yet, because I want to feel these changes out first.

Also I kinda fucked up this PR a bit so theres 2 extra commits that probably don't do anything but I'm still learning github, so.

## Pre-Merge Checklist

## Changelog
:cl:
Buckshot now does .22LR damage per pellet to simplemobs (At time of writing, 15.)
Ratshot and rubbershot, and handloaded buckshot now do handloaded .22LR damage per pellet to simplemobs. (11.25)
Beanbags now do .50BMG Match damage to simplemobs (93.75 or something like that)
AEP-5CR should be doing .38 damage to simplemobs, but I wasn't able to get it to work on a hellpig. Probably would've been a bit much even if it did work!
/:cl:


Tk420634 (approved):
Let's give it a stink

PR #3610, opened 2023-10-27
Loot pool tweaks and trader crate addition

DiscordWizard:
## About The Pull Request

This pull request removes magic weapons from the energy weapons table as to not dillute the available loot too much. Staves and wands can already be crafted by anyone, and their material requirements are easier to get than those of normal craftable energy weapons. The traders also have dedicated crates to buy magical weapons and tools from, it would be silly to have them appear in the normal weapon crates as well.

It also comments out the sord from the unique weapon table because as funny as it was, anyone managing to get their hands into any sort of spawn from the table would probably not find it that hilarious that their chance at a sick gun was instead a Sord.

Adds a very expensive crate (three times the rare crate one) that gives either a single uncommon unique or rare unique weapon once bought. Honestly I low-balled it to see what people would think, I'd prefer it have both the uncommon unique and the rare unique inside because at 15k point price there's absolutely no way to make any sort of benefit as a trader.

Also removes the chance to get an uncommon weapon in your rare crates. It's already 5k points for 3 weapons that together can sell for about half of that, we don't need even more of a chance to have our points wasted with a gun we can get with the cheaper boxes.

## Pre-Merge Checklist

## Changelog
:cl:
add: Added Unique weapons crate to trader
del: Removed sord from unique loot table
del: Removed staves and wands from energy loot tables
del: Removed the 25% chance to roll an uncommon in the 3 items of the rare weapons crate
/:cl:


Tk420634:
I'll pass this one to the SHADOW COUNCIL for deliberation (good hopes imo)

Kelprunner:
So, something to note is the magic items are **not** going to be craftable forever. At least, not without a trait, and the trash tier ones will always be craftable just so people can - no matter what - give them a try. That's about 90% of the reason I added them to the loot pool in the first place. The other 10% is that a fair few people just abjectly refuse to interact with the crafting menu (which is reasonable, fuck TG's crafting menu) and because - at the time - the energy weapon lists were almost as bad as cowboy guns with how little variety there was. Speaking of, I _probably_ should have put the wands there, if not into their own spawner -- but making a new spawner would also require a lot more work since it'd involve someone making a second pass over the map, adjusting the loot spawner chance code, a bunch of math, and so on so forth. As for why they've been craftable for so long, it's a mix of me being ill and having to work on other stuff. Basically, leave the wands in the pool but feel free to move them to Cowboy loot because they're more thematically appropriate there. It's where I'm going to put Weapon Artifacts down the line anyway.

Regarding the shop thing, I'm fine with the shop being able to buy a Unique at stupid high prices, I personally would rather it be Uncommon only. Not _against_ rares being buyable, but we have too much gambling at is it. If you want to insist on making Rare Uniques buyable, split them into two crates: Uncommon Unique and Rare Unique.

DiscordWizard:
Splitting the crates into uncommon and rare uniques sounds good, the point of adding them was to cut down a little on the gambling and that would help even more. I'll just have to adjust the prices a little. 

As for the wands, yeah the cowboy loot makes more sense than the energy weapon loot if artifacts will go there as well, I only hope they don't appear in normal weapon crates since there's already magic crates for them.

Kelprunner:
Yeah, it wasn't my best call putting them in e-weps. Not actually sure what my logic was there beyond "I need to get them in the loot pool."

Tk420634:
> Yeah, it wasn't my best call putting them in e-weps. Not actually sure what my logic was there beyond "I need to get them in the loot pool."


Pet the kelp.  So what's the plan with this then?

DiscordWizard:
Well I moved the staves and wands over to the cowboy loot table and separated the boxes between uncommon unique for 10k and rare unique for 15k, if those values are alright that should be it I think?

Kelprunner:
Sorry for the late reply, I'm terrible about checking github for comments and such and that's on me.

You'll need to, uh, update your PR a little and remove the Magic Missile staff from masterlootdrop.dm because it's bugged to high heavens right now. As for your unique crates, they look good to go? The prices can be tweaked down the line.

DiscordWizard:
There, that should do it

PR #3845, opened 2023-11-19
Reduces range of intercontinental ballistic merp

Superlagg:
## About The Pull Request
Now you gotta be in the same room with someone to see them chokeslam their hog.


Tk420634 (approved):
oink oink, I guess

PR #3938, opened 2023-11-27
A real CDDA...

Kilmented:
## About The Pull Request
Adds garter belt rigs to the loadout, as an item, and as a crafting recipe.
(not at all for a robomaid shhshh stop asking questions and using your brain just SPEEDMERGESPEEDMERGESPEEDMERGESPEEDMERGE)

## Why is this good for the game?
Because you can re-name it and just use it as a grenade rig or a throwing knife rig or just as more pockets. Who doesn't love more pockets? I do. I love pockets. Hehe. They're so warm and I like taking things from them. Hehe. 

## Pre-Merge Checklist

(I'm testing the PR as we speak so if you can read this you might not want to merge just yet but it's super harmless and there's no icons or anything to worry about so like just merge it anyways) 

## Changelog
:cl:
add: new belt slot item for holding grenades in a tacti-lewd way or maybe not if you just rename it in the loadout menu I don't judge I am not your mom.
/:cl:

### DIFF
--- a/code/datums/components/crafting/recipes/recipes_clothing.dm
+++ b/code/datums/components/crafting/recipes/recipes_clothing.dm
@@ -86,6 +86,10 @@
 	category = CAT_CLOTHING
 	subcategory = CAT_GENCLOTHES
 
+/datum/crafting_recipe/legholster/throwing
+	name = "Garter Rig"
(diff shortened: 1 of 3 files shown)


Tk420634 (approved):
You are a pocket.

PR #4018, opened 2023-12-07
Map Bug Fixs: Garland

OliOliOnsiPree:
## About The Pull Request
Majority of these changes are putting floors instead of walls on the windows, and bringing the farm up to a more farming standard then having the silly shovel that removes plots, and removes a rogue mob spawner in the town and removes a loot drop that was in the town and shouldnt of been there, as it was free rare items and to the east of the town and also added a few lights too, so it should be as bright as someone smiling instead of darkness. There was a few rogue little caesers doors around and they were zapped into oblivion below is the major changes. 
image
image

## Pre-Merge Checklist

## Changelog
:cl:
tweak: garland
/:cl:

### DIFF
--- a/_maps/map_files/coyote_bayou/Garland-City.dmm
+++ b/_maps/map_files/coyote_bayou/Garland-City.dmm
@@ -370,6 +370,7 @@
 /turf/open/floor/f13/wood,
 /area/f13/building)
 "apB" = (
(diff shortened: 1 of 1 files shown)


Tk420634 (approved):
ayo lets go

PR #4070, opened 2023-12-17
Prevents ghosts from spawning mobs in nullspace mid-disconnect.

GremlingSS:
## Pre-Merge Checklist

Unironically the first bug I've been really impressed by on this codebase, *hopefully* prevents this from happening again. 
image
Love it. Simple basic network replication issues I encountered during all my uni work this semester simply showing up with BYOND's netcode! Though, it's not their code which is the issue, it's our codebase not being safe for concurrent issues such as that.

If anyone's wondering, basically when a observer leaves the game, the ghost they have is deleted. However with our changes to garbage collection, Fenny nullspaces it, so on that same tick now after the player disconnected on the same frame the spawn mob proc is called, that ghost is in nullspace, and it's src.loc is `null`, so that mob is now instantiated into nullspace.
Simple one line of code to make sure src.loc is never null would easily stop this from happening. (Unless it happens again :X then I'll be darn surprised!!)

### DIFF
--- a/code/modules/mob/mob_transformation_simple.dm
+++ b/code/modules/mob/mob_transformation_simple.dm
@@ -26,10 +26,11 @@
 	if(isturf(location))
 		M = new new_type( location )
 	else
-		M = new new_type( src.loc )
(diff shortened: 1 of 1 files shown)


Tk420634 (approved):
Denied.

PR #4076, opened 2023-12-18
Equalizes Biter, Tailer, Clawer To Match Armblader

EtheoBoxxman:
Makes the strongest variants of the biter, tailer, and clawer trait items do 40 force instead of 30, as the armblader did 40 already yet cost the same.

Also changes sabre toothed biter cooldown mod from 0.7 to 0.8, since that seemed like a typo, given the other two had 0.8


Tk420634 (approved):
Feral meta lets go

PR #4104, opened 2023-12-28
Changes to the snore emote

LordPapalus:
## About The Pull Request
Some people asked for the male snore emote to be not used anymore, as the female snore is way quieter and easier on the ears, I did that. I also put in a second thing people asked for, which is the snrr mimi sound effect.

The new sound effect is under the same emote as the other snore, but has a 1 in 500 chance of being heard, plus both sounds have the same volume, so nobody's ears are damaged.

## Pre-Merge Checklist
- [✓] You tested this on a local server.
- [✓] This code did not runtime during testing.
- [✓] You documented all of your changes.

## Changelog
:cl:
tweak: Removed gender check on the snore emote, now the femalesnore.ogg file will play all the time
add: Added a 1 in 500 chance for the snoremimi.ogg file to play in place of femalesnore.ogg
/:cl:


LordPapalus:
image

Tk420634 (commented):
Just lower the volume of the male snore.  The rest of this is a no go.  Snoring exists mostly to let other players know someone is sleeping nearby.

Colfer1:
@Tk420634 "Snoring exists mostly to let other players know someone is sleeping nearby."
This PR was made for me by a suggestion I made on the discord; The female snoring sound is still played, and still sounds like snoring, but is more immersive for me because it actually sounds like someone snoring, rather than being cartoonishly exaggerated and audible from 20 tiles away. If I read the code correctly, snoring sound shouldn't be removed, still audible, just easier on the ears if this PR was merged, unless I'm missing something?

Tk420634:
It also makes snoring much rarer, which we do not want.  All that really needs to be done is the male snore needs to be put in audacity and turned down slightly.  Its 3 seconds worth of work, literally longer to type this out than it would take me. 🤣 I might just do that later, but right now I'm trying to relax some.

LordPapalus:
Since I don't really want to deal with this, and this is mostly an issue that colfer has, I'll close the PR and let him handle it later on.

PR #4147, opened 2024-01-09
Fixes the bluespace paint stats

PrettyPsychoForAWolf:
Thanks to other devs, this actually works how it should now!

Random stats from -35% to 45%. Might get the best mod or trash, good luck!

Also removed the second bluespace paint. It was brought to my attention that two separate ones were unneeded due to the random variety. They can go on any gun, but some might be better on a specific gun type (for obvious reasons).


OverDriveZ (approved):
And in it goooooes then!

PR #4156, opened 2024-01-11
Complete Gun Mod Randomization

PrettyPsychoForAWolf:
The numbers pull from a select range for every stat that is affected!


PrettyPsychoForAWolf:
This PR should also be easily edittable as well in the future if need be. Some are opposite (like recoil for example needing negative numbers to be positive), but other than that it'll be a simple line change.

Tk420634 (approved):
i swear

PR #4170, opened 2024-01-12
Whoops, take a shot of fabuloso

Superlagg:
## About The Pull Request
Fixes staminartifacts.
Also fixes the reddening on healtifacts


OverDriveZ (approved):
In Dan we trust

PR #4226, opened 2024-01-18
Restricts emoticons to actually functional message modes

Superlagg:
## About The Pull Request
Should no longer proc in deadchat or whatever.


Tk420634 (approved):
🧂 🐈

PR #4227, opened 2024-01-18
{TM ONLY} Disables the PA trait requirement.

Kilmented:
See title.


Tk420634:
atomize it u dorkosaurus

PR #4236, opened 2024-01-19
Enables SSD indicator probably

EtheoBoxxman:
This looks like it should work but I cant test it on my own, alas.


Tk420634:
@EtheoBoxxman no dice homie

PR #4245, opened 2024-01-21
Fixes two specific mobs (Enclave Soldier and Tunneler)

PrettyPsychoForAWolf:
Enclave soldier now uses the gutsy plasma bolt instead of the carbine which COMPLETELY goes through armor. No more of that rediculous gunk. Should be far easier but they're still a difficult mix of boss raider and gutsy.

Gives the regular tunneler a weaker version of the nightstalker venom.

Also adds sounds. They're screams. They scream. Enjoy the nightmare fuel!

https://youtu.be/pglbDy_aEfI

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/f13/fallout_NPC.dm
+++ b/code/modules/mob/living/simple_animal/hostile/f13/fallout_NPC.dm
@@ -219,14 +219,14 @@
 	health = 650
 	melee_damage_lower = 20
 	melee_damage_upper = 47
-	extra_projectiles = 2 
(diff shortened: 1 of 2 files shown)


PrettyPsychoForAWolf:
Forgot one key thing...

PrettyPsychoForAWolf:
Mergeable now. Replaced /obj/item/projectile/plasmacarbine with /obj/item/projectile/f13plasma/scatter on the enclave soldier. Should make those fights far more manageable without funny 100 wound damage instantly.

Tk420634 (approved):

PR #4298, opened 2024-01-30
Shopkeeper has better cells

PrettyPsychoForAWolf:
Technophreak from jobs is just broken. No identifiable cause. This'll be a bandaid fix


Tk420634 (approved):
*pulls it off cutely

PR #4301, opened 2024-01-30
Job number fixes

PrettyPsychoForAWolf:
Removes Quartermaster

Adds an extra slot to science, doctor, and trade worker


OverDriveZ (approved):
Quartermaster? More like pennymaster, gotem

PR #4322, opened 2024-02-01
Adds the charger box to normal game

Superlagg:
## About The Pull Request
and maybe breaks autolathes forever, cus course it does


Tk420634 (approved):
As you do.

PR #4470, opened 2024-02-21
Ody licks the den

OdysiaHallow:
See title. I forgot to take screenshots, ma bad.


_maps/map_files/coyote_bayou/Nash_and_Texarkana-Upper.dmm
Tk420634, line 1: Well, you can.  But do /".

So

desc = "For cleaning up /"messes/", you suppose. It smells terrible";

would have been fine to.

Tk420634 (approved):

PR #4474, opened 2024-02-21
Re-enables charge cost reduction/increase in mods and Adjusts Damage Numbers

PrettyPsychoForAWolf:
Forgot to bring the battery shunt in the mix. Also, it among other things have charge cost increase now. Its a *battery shunt*. No shunting was done originally.


Tk420634 (approved):
Imma shunt ur legs

PR #4520, opened 2024-02-25
Makes the snow thing tied to ssweather

Superlagg:
ssssssssssweather.

also turns off snow early cus pandimensional global warming.


Superlagg:
turns out its already in the game, sick

PR #4597, opened 2024-03-06
Gives Guild access credentials to Scientists

PrettyPsychoForAWolf:
Only for upgrades/tool delivery.

Can be used to be a doctor if no doctor is around.

Plx bap anyone that abuses, thanks!


Tk420634:
I'd rather not encourage people to not interact, sorrrrryyy blue

PR #4618, opened 2024-03-11
I HATE FUCKING PAPER

Aphast:
## About The Pull Request

Disables report card printing, as well as tracking.

Tested on a private server using multiple 180 mags worth of bullets with both a reloading bench and autolathe no report printed, mats were still added, and upgrades were able to be applied. No signs of printing paper either.
## Pre-Merge Checklist
- [ X] You tested this on a local server.
- [outside of normal ones] This code did not runtime during testing.
- [X ] You documented all of your changes.

## Changelog
code: comments out all report card systems
/:cl:

### DIFF
--- a/code/game/machinery/autolathe.dm
+++ b/code/game/machinery/autolathe.dm
@@ -47,10 +47,10 @@ GLOBAL_VAR_INIT(lathe_reports_done, 0)
 	var/DRM = 0 //determines if DRM is active, preventing you from making ammo
 
 	/// If someone is loop-feeding us, let em know what they put in when it stops looping
-	var/datum/autolathe_loop_returns/report_card
-	var/report_timer
-	var/datum/autolathe_loop_returns/big_report_card
-	var/big_report_timer
+//	var/datum/autolathe_loop_returns/report_card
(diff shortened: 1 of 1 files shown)


Superlagg:
If you do this, you'll need to make them tell the player what they recycled and for how many materials.

Aphast:
From questionares about, apparently noone reads those in the first place, there is also the simple option of just clicking 'show mats' before and after

Superlagg:
So I kinda like the mess it makes, and that was the original intent, to make a mess. What I *could* do is have the paper pass through things within 1-2 tiles of the machine so that it doesnt block the bench.

Aphast:
eh, if ya like it sure w/e, but the issue on my part arises from it flinging shit halfway across the entire town instead of just ruining the workshop, however an idea WAS suggested to make it syndi origami or have increased damage, of which I am clueless as to how to code

Superlagg:
yeah I could make the paper be syndie origami, give it some use aside from making a mess.

Aphast:
best of luck with such then. no need for this PR

PR #4633, opened 2024-03-13
Fixing toxins hunger processing on mechs

Colfer1:
## About The Pull Request

This should fix toxins processing and hunger on mechs. The reason why I added easy dismember instead of nodismember is that I don't think its possible to dismember people regardless of the traits, and I wanted to make it as close as possible to synthliz.dm so that sleep healing hopefully works on mechs. 
!! I cannot test these changes, because my test server really hates the adapted species, and they do not appear regardless of what I do !!
## Pre-Merge Checklist

## Changelog
:cl:
add: Added new things
add: Added more things
del: Removed old things
tweak: tweaked a few things
balance: rebalanced something
fix: fixed a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
server: something server ops should know
/:cl:

### DIFF
--- a/modular_coyote/code/creature_species.dm
+++ b/modular_coyote/code/creature_species.dm
@@ -1196,7 +1196,7 @@
 	icon_dead_suffix = "_dead"
 	icon_rest_suffix = ""
 	species_traits = list(FERAL,NOZOMBIE,NO_UNDERWEAR,MUTCOLORS,NOTRANSSTING,EYECOLOR,ROBOTIC_LIMBS,NO_DNA_COPY,NOEYES,LIPS,)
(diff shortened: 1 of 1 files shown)


Colfer1:
This won't have a guaranteed chance of fixing sleep healing, but it'll atleast make robots not process oxygen/toxin damage like oxygen damage from crit, and permanently have their hunger set to max. No idea if this will work, and if it doesn't allow sleep healing, I'm not a powerful enough coder to make it work. May make robots lose their limbs, in which case I don't know what to do about that

Colfer1:
Still working on this, will comment again when finished

Colfer1:
Should be good assuming all checks pass

Tk420634 (approved):

PR #4650, opened 2024-03-16
Second wind mulligan

Superlagg:
## About The Pull Request
If you die again within 5 minutes of second winding, you get a free life! So if your second wind fails for some reason, you can just try (and fail) again!


Varoxus:
that summary is reminding me of the meat grinder rant.

Tk420634 (approved):

PR #4743, opened 2024-03-29
Removes vestigial lead

Aphast:
## About The Pull Request

Removes the material and reagent known as 'lead'(leadacetate is unaffected as it is actually not lead). Unlike other materials that are properly interspersed throughout the proto and autolathe fab list, lead is entirely useless due to being a vestigial material no longer in use. Even for the alloy recipe(changed to iron) it doesn't work due to some unknown bug, and lead cannot be put back into the ORM. ATM lead just serves to be an annoyance with no real purpose outside of the niche use of what anomalous cubes are pathed from and reducing mining yields by roughly 10% or so(im not doing the full math im sick of stats). However they have been changed to instead default metal (will never actually happen due to them only being made by players, and metal already existing). 

Private server testing shows alloy can be crafted, as well as anomaly boxes, with no clear runtimes either. (iron and gold pictured alongside alloy made in the ORM, the protolathe also works)
image

## Pre-Merge Checklist

## Changelog
:cl:
del: All lead paths and mentions
balance: Alloy now requires iron instead of lead
/:cl:

### DIFF
--- a/_maps/map_files/Pahrump-Sunset - Backup/RedRiver.dmm
+++ b/_maps/map_files/Pahrump-Sunset - Backup/RedRiver.dmm
@@ -8595,7 +8595,6 @@
 	},
 /area/f13/building)
 "eTF" = (
-/obj/item/stack/ore/lead,
 /turf/open/water,
 /area/f13/building/abandoned)
 "eUh" = (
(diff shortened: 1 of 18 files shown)


Tk420634 (approved):
industrial society moment

PR #4776, opened 2024-04-03
nanite tweaks

Kilmented:
nanites (can) have literally 0 downside

alternatively, for even more nanites, you can be slightly hungrier

tweaks the abilities to be slightly more powerful- at a severe cost. You can't have them all active as passives anymore.


Tk420634 (approved):
get them to make people big and round

PR #5005, opened 2024-04-21
NB Rework WIP DO NOT MERGE

OdysiaHallow:
This is an in progress rework of NB. Do not merge it yet.

Yes I am aware the spawns aren't down, that things are missing, and there are no lights. This is a WIP. Feel free to let me know if you have any ideas or suggestions!


Aphast:
two small things, please do not forget a hydroponics grinder, and when it comes to ORM placement, both vendors should ideally be next to the ORM, instead of opposite sides of town

PR #5088, opened 2024-04-30
BOTANY BROS WE ARE SO BACK

Aphast:
## About The Pull Request

image
Finally fixes the damn trays not having a max of 100, total and max are not actually the same, which max is what was altered during the previous time of parts actually working on trays. Now instead of defaulting to 20 it will properly stay stuck at 100, with a bonus of 100 nutrient roundstart. Also removes bugged trays from NB, and adds the vendor closer to the ORM that fenny forgor. Unsure if the drain is still horrid, but maybe it will be better now, perhaps a botanist thread on the discord can help with keeping track of times, for now everyone gets to keep rushing gaia.
## Pre-Merge Checklist

image

## Changelog
fix: reagents.max_volume null refrence
add: mining vendor next to the orm like a sane person
fix: building south of NB no longer spawns bugged circ boards
/:cl:

### DIFF
--- a/_maps/map_files/coyote_bayou/Newboston.dmm
+++ b/_maps/map_files/coyote_bayou/Newboston.dmm
@@ -13372,6 +13372,10 @@
 "gvH" = (
 /turf/open/floor/carpet/purple,
 /area/f13/building/abandoned)
+"gvN" = (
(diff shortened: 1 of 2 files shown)


Tk420634 (approved):
Denied because you smell really bad

PR #5127, opened 2024-05-03
Reload Gun Hotkey

TalkingCactus:
Press 1 button to reload the gun in your active hand, assuming it uses external magazines. Can be adapted later to work with toploaders/shotguns/energy weapons/etc. It will search 1 layer deep into any pockets, bags, rigs, armor, shoes, helmets, etc that you might be wearing. Has trouble with magazines that are stored in your pockets and stuff but aren't in another item.

Takes 1 second to reload by default, but it can be altered on a gun-by-gun basis with a single variable. A variable on magazines would be a good idea too, but that can come later if we feel we need it for balance.

When you use it, it will replace the magazine that you pulled out with the empty one you just ejected if possible. Otherwise, it's ejected to the floor.

### DIFF
--- a/code/_onclick/hud/screen_objects/character_actions.dm
+++ b/code/_onclick/hud/screen_objects/character_actions.dm
@@ -87,11 +87,11 @@
 				H.emote("tend")
 
 			I = H.get_active_held_item()
-			I.melee_attack_chain(H, H, params)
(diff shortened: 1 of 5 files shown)


Tk420634:
image

Tk420634 (approved):

PR #5295, opened 2024-06-29
DT pierce return

Kilmented:
compact gets five
med 10
long 15
shotguns get 5, but they have 4 pellets, so a total of 20


Superlagg:
oh, damage_flat is only used if no other damage options are set. This thing currenly doesnt support DT pierce, but its an easy add i can do,

Kilmented:
mb

PR #5352, opened 2024-09-22
Ports Visualchat (and some other stuff that was stuck to it)

Superlagg:
This ports visualchat from the bar. It also ports the speech bubble changes and the talk distance thing. needs testing!


Tk420634:
Daaaan if you can't find the linter failure can we please just meeerge thiiiissssss

Tk420634 (approved):

PR #5418, opened 2024-10-01
Removes resctictions from some storage items

Fragenstien:
## About The Pull Request
Removes restrictions from some storage items. More specifically, as many as I thought I could get away with.
Notable exceptions:

- Bags and other "gather all" type containers: Skipped to preserve their functionality
- Boxes: The weird ones, like cigarette boxes and monkey cube boxes. These ones have the storage capacity of a normal box, but may be a different size or have other functionality. Might be a project in the future to remove the restrictions and change their storage size to make sense at the same time, but that's outside the scope of this PR.
- Back Sheath: Seems like it's intended to both provide a sheath for weapons and as a backpack. Without the restrictions it just becomes a backpack within a backpack.
- Sturdy Quiver: Because spears and javelins are so big, the storage capacity on this thing is huge to accommodate carrying multiples.
- Everything in the storage/fancy category. These are containers that have different icon states depending on what's in them. These are left unchanged for the sake of clarity.

One notable change is to portable rechargers. Specifically, you can now replace the capacitor in the internal box with anything. If it's not a capacitor, it doesn't work. I was afraid messing with this would break something, but it all seems okay.

## Pre-Merge Checklist

### DIFF
--- a/modular_sunset/code/mob/renegade.dm
+++ b/modular_sunset/code/mob/renegade.dm
@@ -31,7 +31,7 @@
 	waddle_amount = 5
 	waddle_up_time = 1
 	waddle_side_time = 1
-	retreat_health_percent = 0.5
+	retreat_health_percent = 0.05
 	max_heal_amount = 0.9
 	heal_per_life = 0.115
 	tactical_retreat = 30


Tk420634:
I'm not sure what you did here homie, but this pr is undoing a ton of changes that we just recently did.  So uh.  No.  Lets get the things you actually want cherry picked out of it. lol

PR #5572, opened 2025-01-08
moves a bunch of HUD buttons into a radial (two radials)

TalkingCactus:
that'll be three fiddy
images
also fenny needs to make icons for the buttons before this is ready


Tk420634 (approved):
get in voice we hoonting in a bit

